PR #40, opened 2019-01-25
Rock-2464: Create scss to move Employment page to Rock

mhamner:
Create new scss partial for Employment page and update init.scss to import the new partial.  This is for the move of the Employment Opportunities page to Rock.


ColoradoWebGuy:
Looks great, thanks!

PR #59, opened 2019-07-18
theme edits

bengoshow:
photo cropping for directory
event accordion wrapper
self-development content channel
correction to video link


bengoshow:
@cbump can you merge these updates into the repo and the live theme? Thx

PR #66, opened 2019-09-13
prep files for handoff

bengoshow:
Final style tweaks.
Removed SECC2019 Theme imports and just linked to those compiled css files directly in Site.Master for a much smaller seportal.css file


bengoshow:
@cbump made one last tweak before handoff. Let me know if you need anything else from me.

PR #77, opened 2020-01-30
Add transformations to recurring communications.

chrisfunk:
Add the ability to perform person transformations on a recurring communication data view.  

An example of this is if a user wanted to email a child's parents a Happy Birthday card on their birthday.


trydyingtolive:
Thanks Chris!

PR #154, opened 2024-09-10
Upgrade 1.13.7 datably

JulioCachay:


datably/scripts/README.MD
```diff
@@ -0,0 +1,28 @@
+﻿# Scripts to run V1.13.7
+## Overview
+The main motivation for these scripts is to facilitate local development so we can more efficiently:
```
datably, line 3: "so we can ____ more efficiently:"

Plugins/org.secc.PastoralCare/org.secc.PastoralCare.csproj
```diff
@@ -89,6 +86,10 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\Rock\Rock.csproj">
```
cfunkdev, line 89: can we update this reference to use '$(SolutionDir)\Rock\Rock.csproj' that would help with the Azure Pipleline that SECC uses to deploy rock updates to production

cfunkdev: @JulioCachay the PR looks good, this would be the only update that I would suggest.

cfunkdev (approved):

PR #185, opened 2025-12-16
ROCK-8112 Reduced the amount of check-in cache rebuilds

stphnlee:
The cache was running UpdateKeys too often instead of relying on AllKeys.


stphnlee:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/secc/RockPlugins/pull/185#pullrequestreview-3584944699)

PR #216, opened 2026-04-15
ROCK-8407 Disable/restore communications for inactive people

jwakefield-secc:
## Summary
- Adds `DisableCommunicationsForInactivePeople` job — snapshots and disables `EmailPreference` + per-phone `IsMessagingEnabled` for people whose record status is Inactive. Supports configurable lookback days and Dry Run.
- Adds `RestoreCommunicationsForReactivatedPeople` job — restores preferences from the snapshot for anyone who has gone back to Active, then removes the tracking entry. No lookback so missed runs don't strand anyone.
- Snapshots live in a new `Inactive Person Communication Overrides` DefinedType (Value = PersonId, Description = JSON snapshot).

Ticket: https://seccdev.atlassian.net/browse/ROCK-8407

## Rollout notes
The initial backfill (~307k email flips, ~48k SMS phone flips on prod-sized data) is handled by a one-time SQL setup script run directly against the database — not by this job. The jobs are designed for ongoing incremental work (default 15-day lookback on the Disable job).

## Test plan
- [x] Build the solution — no compile errors
- [x] Deploy to DEV
- [x] Run the one-time setup SQL on DEV (creates the DefinedType and bulk-processes existing inactive people inside a transaction with dry-run counts and 5 sample snapshots for review)
- [x] Spot-check 3–5 tracking DefinedValues on DEV — confirm JSON parses and phone IDs match
- [x] Pick a test person, flip to Active, run Restore job (Dry Run → real) — confirm preferences restored and tracking entry deleted
- [x] Flip a different test person to Inactive, run Disable job with `LookbackDays=15` (Dry Run → real) — confirm snapshot + disabled preferences
- [x] Run Disable job twice in a row — second run should be a no-op (idempotency)
- [x] Schedule both jobs at non-overlapping times (2am Disable / 3am Restore) and let them run overnight on DEV — no exceptions in Exception Log
- [ ] Promote to PROD once DEV is stable

### DIFF
--- a/Plugins/org.secc.Jobs/DisableCommunicationsForInactivePeople.cs
+++ b/Plugins/org.secc.Jobs/DisableCommunicationsForInactivePeople.cs
@@ -1,0 +1,213 @@
+// <copyright>
+// Copyright Southeast Christian Church
+//
+// Licensed under the  Southeast Christian Church License (the "License");
+// you may not use this file except in compliance with the License.
+// A copy of the License shoud be included with this file.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Data.Entity;
+using System.Data.Entity.SqlServer;
+using System.Linq;
+
+using Newtonsoft.Json;
+using Quartz;
+using Rock;
+using Rock.Attribute;
+using Rock.Data;
+using Rock.Jobs;
+using Rock.Model;
+using Rock.Web.Cache;
+
+namespace org.secc.Jobs
+{
+    /// <summary>
+    /// Job that disables email and SMS communication preferences for people
+    /// whose Record Status is Inactive. Stores original preferences in a
+    /// DefinedType so they can be restored if the person becomes active again.
+    /// </summary>
+    [DisplayName( "SECC | Disable Communications for Inactive People" )]
+    [Description( "Disables email and SMS preferences for people with an Inactive record status. Original preferences are stored in a DefinedType for later restoration." )]
+    [DefinedTypeField( "Inactive Person Communication Overrides Defined Type",
+        Description = "The Defined Type used to store original communication preferences for inactive people.",
+        IsRequired = true,
+        Key = "InactiveCommunicationOverridesDefinedType",
+        Order = 0 )]
+    [IntegerField( "Lookback Days",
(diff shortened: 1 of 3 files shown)


Plugins/org.secc.Jobs/DisableCommunicationsForInactivePeople.cs
```diff
@@ -0,0 +1,213 @@
...
+                            IsMessagingEnabled = ph.IsMessagingEnabled
+                        } ).ToList()
+                    };
+
+                    // Store original preferences as a DefinedValue (Value = PersonId)
+                    var definedValue = new DefinedValue
+                    {
+                        DefinedTypeId = definedType.Id,
+                        Value = person.Id.ToString(),
+                        Description = JsonConvert.SerializeObject( snapshot ),
```
gmcgrady, line 179: Great use of the Defined Value's description field to store the preferences snapshot as JSON!

Plugins/org.secc.Jobs/DisableCommunicationsForInactivePeople.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+
+            int disabledCount = 0;
+            int errorCount = 0;
+
+            foreach ( var person in inactivePeople )
+            {
+                try
+                {
+                    var personContext = new RockContext();
```
gmcgrady: Instantiating a new RockContext on every iteration of the loop is good practice, but performance could be optimized by implementing the using statement pattern ( i.e., `using (var rockContext = new RockContext())` ) as a block that encases all subsequent logic. This ensures that disposal & SQL disconnection occurs after every iteration of the loop. Otherwise, that instance is left in memory until C#'s garbage collector gets around to cleaning it up.

Plugins/org.secc.Jobs/RestoreCommunicationsForReactivatedPeople.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+        }
+
+        public virtual void Execute( IJobExecutionContext context )
+        {
+            JobDataMap dataMap = context.JobDetail.JobDataMap;
+            var definedTypeGuid = dataMap.GetString( "InactiveCommunicationOverridesDefinedType" ).AsGuid();
+            bool dryRun = dataMap.GetString( "DryRun" ).AsBoolean();
+
+            var rockContext = new RockContext();
```
gmcgrady, line 78: Performance could be optimized by implementing the using statement pattern ( i.e., using (var rockContext = new RockContext()) ) as a block that encases all subsequent logic. See earlier comment for explanation.

gmcgrady (changes requested):
Overall, solid implementation! Per comments, would recommend a `using` statement block in place of `var personContext = new RockContext();` on each iteration.

gmcgrady (approved):
Improvements made -- looks great! Approving & merging.

PR #220, opened 2026-05-01
ROCK-8488 Fall back to minor's mobile phone when parent phone is blank

jwakefield-secc:
## Summary

Three improvements to the Decision Analytics report (page 4657) so staff can reach more candidates and navigate the report more easily.

## Changes

- **Phone fallback for minors** — when a minor's ParentPhone is blank but they have their own MobilePhone on file, surface their own number in the grid instead of leaving the cell blank. (`DecisionReportItem.cs`)
- **Profile-link icon column** — leftmost column with a clickable user icon linking to `/Person/{id}` so staff can jump to a candidate's full Rock profile in one click. Excluded from Excel export via `ExcelExportBehavior=\"NeverInclude\"`; click does not bubble to the row-select handler. (`DecisionAnalytics.ascx`)
- **Modal bug fix** — for minors, the modal's \"Parent Email\" literal was bound to \`decision.Email\` (candidate's own email) instead of \`decision.ParentEmail\`. Now shows the parent's email under the correct label. (`DecisionAnalytics.ascx.cs`)

Screenshot 2026-05-04 154155

## Test plan

- [x] Open page 4657 — confirm a user-icon column appears as the leftmost column on every row
- [x] Click an icon — should navigate to \`/Person/{id}\` (no modal flash)
- [x] Open a minor's row modal — \"Parent Email\" label shows the parent's email, not the candidate's
- [x] Find a minor with blank ParentPhone but populated own MobilePhone — Phone column shows their number, not blank
- [x] Export grid to Excel — icon column is absent from the export

### DIFF
--- a/Plugins/org.secc.Reporting/Model/DecisionReportItem.cs
+++ b/Plugins/org.secc.Reporting/Model/DecisionReportItem.cs
@@ -175,7 +175,7 @@ namespace org.secc.Reporting.Model
             {
                 if (IsMinor)
                 {
-                    return ParentPhone;
+                    return ParentPhone.IsNotNullOrWhiteSpace() ? ParentPhone : MobilePhone;
                 }
 
                 return MobilePhone;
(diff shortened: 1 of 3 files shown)


gmcgrady (approved):
Code appears ready to push. Approving & merging. Will need to let Steven Estes know about this implementation, if he's not already aware.

PR #221, opened 2026-05-07
Bootstrap org.secc.Test xUnit project + wire into v16 pipeline

stphnlee:
## Summary

Phase 1 of the testing initiative described in the local plan file.

Creates `Plugins/org.secc.Test/` — an xUnit test project that builds alongside every other plugin via the existing `Build Plugins` ForEach in `Rock v16.yml`. Wires a `VSTest@3` step into the same pipeline so the build fails if any test fails.

This PR contains **no production code changes** — `org.secc.Test` is a brand-new project with one smoke `[Fact]` (`SmokeTests.TruthIsTrue`) so we can demonstrate end-to-end that the harness works before extracting any real logic.

## What's in the box

- `Plugins/org.secc.Test/org.secc.Test.csproj` — net472 class library, Rock DLL refs via `$(SolutionDir)RockWeb\Bin\` HintPaths, `ProjectReference` to `org.secc.Finance` (the Phase 2 pilot target), **empty `<PostBuildEvent>`** so test DLLs are NOT copied into RockWeb\Bin.
- `Plugins/org.secc.Test/packages.config` — xUnit 2.9.2, xunit.runner.visualstudio 2.8.2, Microsoft.NET.Test.Sdk 17.11.1, Moq 4.20.72, FluentAssertions 6.12.2 (NOT 7.x — license changed), MockQueryable.Moq 7.0.3.
- `Plugins/org.secc.Test/App.config` — subset of Finance's binding-redirect block (Newtonsoft, System.Memory, System.Buffers, System.Runtime.CompilerServices.Unsafe, etc.).
- `Plugins/org.secc.Test/Properties/AssemblyInfo.cs` — modelled on `org.secc.Mapping/Properties/AssemblyInfo.cs`.
- `Plugins/org.secc.Test/SmokeTests.cs` — single `[Fact] TruthIsTrue()`.
- `Rock v16.yml` — adds `VSTest@3` step between `Build Plugins` and `Copy Plugin Files to Artifact Staging`, with `failOnMinTestsNotRun: true` so a missing test DLL fails the build. (This file was previously untracked locally; the commit captures it as well as the new step.)

## Test plan

- [ ] Pipeline build succeeds on this branch's merge into `hotfix-1.16.12`.
- [ ] DevOps test results tab shows 1 passing test (`SmokeTests.TruthIsTrue`).
- [ ] `RockWeb.zip` artifact does NOT contain `bin/org.secc.Test.dll` (verify by extracting the zip and grepping `bin/`).
- [ ] Sanity-check the failure path: temporarily change `true.Should().BeTrue()` to `false.Should().BeTrue()` on a follow-up branch, push, confirm pipeline goes red. Revert.

## Notes for reviewer

- The `Build Plugins` ForEach loop will pick up `org.secc.Test.csproj` automatically and run `msbuild`. The xcopy fallback in that loop only activates for plugins with an `org_secc/` subfolder — `org.secc.Test` has none, so it's safe.
- The pipeline cleanup task that retains `RockWeb`/`packages`/`libs` runs **before** the plugin clone, so the new VSTest step (which runs after `Build Plugins`) reads from `$(Build.SourcesDirectory)\RockPlugins\Plugins\org.secc.Test\bin\**\org.secc.Test.dll` while that folder still exists.
- No real plugin logic is touched in this PR. Phase 2 (extract `Utility/Statement.cs::AddMergeFields` from `org.secc.Finance` into `Services/StatementService.cs` and add real tests) lives in a follow-up PR.


stphnlee:
Addressed Copilot's review in `3a65d08b`.

**Fixed (real issues):**
- **#3 — Clean wiping the test DLL.** The `Build Plugins` loop ran `msbuild -target:Clean` immediately after build, which would have deleted `org.secc.Test\bin\Debug\` before VSTest could find the assembly. Now skips Clean specifically for `org.secc.Test`.
- **#4 / #7 — VSTest discovering zero tests.** Classic packages.config csprojs do not auto-import the xUnit + Microsoft.NET.Test.Sdk build props/targets, so without explicit `<Import>`s the VS test adapter never lands next to `org.secc.Test.dll`. Added the imports NuGet would normally write for a packages.config test project (xunit.core props/targets, xunit.runner.visualstudio props, Microsoft.NET.Test.Sdk props/targets, xunit.analyzers props) plus an `EnsureNuGetPackageBuildImports` target that errors clearly if any are missing.
- **#6 — Non-deterministic restore.** packages.config now lists `EntityFramework`, `Newtonsoft.Json`, `System.Runtime.CompilerServices.Unsafe`, and `System.Threading.Tasks.Extensions` explicitly, instead of free-riding on whichever plugin happens to restore them first.

**Pre-existing in the v16 YAML, not introduced by this PR — flagging for follow-up:**
- **#1** — Job pool only specifies `name: Azure Pipelines` with no `vmImage`. Likely fine because of how the pool is configured in DevOps, but worth a follow-up.
- **#2** — The `Cherry-Pick Core Changes` PowerShell task has both `targetType: inline` AND a `filePath` input. The `filePath` value is actually a verbatim copy of the script body, so YAML validation should still pass and `script:` is what runs at execution time, but it's confusing and worth cleaning up.
- **#5** — `AspNet.Security.OpenIdConnect.Primitives.dll` is copied twice in the Add-Package-Dependencies PowerShell.

These three pre-date this PR (the v16 YAML was previously untracked locally). Out of scope for the bootstrap PR but happy to follow up in a separate housekeeping PR if you want.

stphnlee:
All three findings in this round are pre-existing in the v16 YAML (none touched by this PR's diff except via the unrelated insertion of the new VSTest step). Already covered in the previous response — out of scope for the bootstrap PR, but happy to follow up in a separate housekeeping PR.

For the record:
- **Cherry-Pick `targetType: inline` + `filePath`** — same as prior #2. The `script:` field is what runs at execution time, but the redundant `filePath:` should be removed for clarity.
- **`$plugin` DirectoryInfo coercion in `Build Plugins`** — same shape as prior #2 in the v13 file. The pipeline has been building Rock + every plugin successfully in production with this script, so PowerShell's coercion is producing usable paths in practice; still worth tightening to `$plugin.Name` or `Join-Path` defensively.
- **Pool missing `vmImage`** — same as prior #1.

If you want a single follow-up PR cleaning these up alongside the duplicate copy line (#5), happy to open one once Phase 1 is merged and green.

PR #222, opened 2026-05-07
Updating server health block to only monitor for maintenance flag

stphnlee:
This block was doing a lot of complicated calculations of message bus usage and monitoring health of other servers, but it was actually causing additional reliability issues. This update restores the maintenance flag functionality without the additional complexity.


gmcgrady (approved):
Per discussion, makes sense that we would simplify this to only rely on the maintenance flag for a logical "unhealthy" state determination. Otherwise, if the gateway can reach this page, then we know it's healthy. Approving & merging.

PR #224, opened 2026-05-11
ROCK-8501 Add MedicationActive flag and source-aware medication manager

jwakefield-secc:
## Summary

Restores safe medication management on the camp medication dispense page by differentiating behavior per data source:

- **Snapshot camps** (most summer camps — registration workflow copies meds to a GroupMember-level matrix): full Edit / Delete / Add Medication on the snapshot. Edits stay scoped to the event.
- **Person-master camps** (HSM Fall Retreat, MSM Discipleship Weekend, RUCKUS, etc. — no workflow): only Activate / Inactivate via a new `MedicationActive` boolean attribute on the matrix template. Inactivating skips the med for the event without touching the master record.

Replaces the view-only modal from the prior ROCK-8328 follow-up (`01273c59`), and supersedes PR #209.

## Setup required per environment

Add a new attribute on the medication matrix template (Guid `d2ed9b3d-309a-4a70-bda4-2c090acd1384`):

- **Name:** `Active`
- **Key:** `MedicationActive`
- **Field Type:** `Boolean`
- **Default Value:** `True`

Either via the Rock UI ("Template Item Attributes" section) or via SQL insert. The new migration `006_MedicationReportFilterInactive.cs` references this attribute by key at runtime so the stored proc stays portable across environments.

After adding, **clear Rock cache** so the running app picks up the new attribute metadata.

## Code changes

| File | Change |
|---|---|
| `MedicationDispense.ascx` | Modal grid: source-badge, Active+Toggle columns, Edit/Delete columns, Add Medication button, Edit sub-modal |
| `MedicationDispense.ascx.cs` | Grid query restructured (`Concat` of two simple INNER-JOIN queries — snapshot path + Person fallback via `NOT EXISTS` antijoin — replacing an earlier `GroupJoin + CASE` approach that caused SQL timeouts). Source detection via `IsSnapshotMatrix`. Toggle / Edit / Delete / Add handlers, all gated to the correct source. Skip-inactive in dispense loop. |
| `GroupLeaderMedicationNotifications.cs` | Notification job's matrix-item join LEFT-joins `MedicationActive` and skips items with `Value = "False"`. Treats absent values as active. |
| `MedicationInformation.ascx.cs` | New `FilterInactiveMatrixItems` helper called from `BindLava` and `BindGrid` so family-facing view hides inactive meds. |
| `CreateMedicationLabels.cs` | Kiosk check-in skips inactive items when generating medication labels. |
| `Migrations/006_MedicationReportFilterInactive.cs` (new) | Re-creates `_org_secc_CampManager_GetMedicationReport` stored proc with a `LEFT JOIN` on `MedicationActive` and `WHERE [Value] IS NULL OR [Value] != 'False'`. Looks up the attribute Id at runtime by key + template-Guid qualifier. |

All consumers use `AsBoolean(true)` semantics — matrix items with no stored `MedicationActive` value are treated as active.

## Test plan

- [x] **Phase 1 — toggle Active/Inactive on no-snapshot camp.** Verified on HSM Fall Retreat (page 2818): toggle writes `False` to the AV, med disappears from dispense grid, reactivates cleanly.
- [x] **Phase 2 — Edit / Delete / Add Medication on snapshot camp.** Verified on HSM Camp (page 1389) with a cloned snapshot. All three actions work and the Person master matrix is unchanged after snapshot edits.
- [x] **Grid query performance.** No SQL timeouts on snapshot camps after the `Concat` restructure.
- [x] **`MedicationInformation` block (`/myevents`).** Family-facing view hides inactive meds end-to-end.
- [x] **`GroupLeaderMedicationNotifications` job — filter predicate verified by SQL.** Job logic exactly matches our `IS NULL OR != 'False'` predicate; TEST ADVIL on test person classifies as EXCLUDED. Job-trigger test deferred (dev env has no active reminder-job descendant groups + concern about real comms). The join chain itself ran successfully 18 times in 2025 production cron runs — our change is just a `LEFT JOIN + WHERE` inserted into that working chain.
- [x] **`CreateMedicationLabels` (Medication Check-In workflow #191) — predicate verified by SQL.** Same `AsBoolean(true)` pattern as the verified family-facing block; same data, same EXCLUDED result. "All inactive" case correctly routes to the existing "No Medication Information Found" label fallback.
- [x] **Camp medication report stored procedure — end-to-end verified.** Ran `EXEC dbo._org_secc_CampManager_GetMedicationReport @RegistrationTemplateId = 1483` (2025 HSM Camp template). Test person rows returned `Tylenol` and `Methaline Blue` but **not** `TEST ADVIL` — the inactive med is correctly excluded from the Excel export pipeline.

## Open question (not blocking)

Should `/myevents` allow parents to toggle Active/Inactive directly, or stay view-only? See the [open question comment](https://github.com/secc/RockPlugins/pull/224#issuecomment-4421989261) for the trade-offs. Currently view-only; no change in this PR.

### DIFF
--- a/Plugins/org.secc.FamilyCheckin/Workflows/CreateMedicationLabels.cs
+++ b/Plugins/org.secc.FamilyCheckin/Workflows/CreateMedicationLabels.cs
@@ -131,7 +131,17 @@ namespace org.secc.FamilyCheckin
                                 mergeObjects.Add( "People", people );
                                 mergeObjects.Add( "GroupType", groupType );
 
-                                if ( attributeMatrix == null || attributeMatrix.AttributeMatrixItems.Count == 0 )
+                                // Filter inactive medications so the kiosk doesn't print labels for meds the leader has marked inactive.
+                                var matrixItems = attributeMatrix?.AttributeMatrixItems
+                                    .Where( ami =>
+                                    {
+                                        ami.LoadAttributes();
+                                        return ami.GetAttributeValue( "MedicationActive" ).AsBoolean( true );
+                                    } )
+                                    .ToList()
+                                    ?? new List<AttributeMatrixItem>();
+
+                                if ( attributeMatrix == null || matrixItems.Count == 0 )
                                 {
                                     // Add a No Medication Information label for anyone without data
                                     var checkInLabel = new CheckInLabel( labelCache, mergeObjects );
@@ -149,7 +159,7 @@ namespace org.secc.FamilyCheckin
                                 }
                                 else
                                 {
-                                    var items = attributeMatrix.AttributeMatrixItems.ToList();
+                                    var items = matrixItems;
                                     var index = 0;
 
                                     while ( index < items.Count )
--- a/Plugins/org.secc.Jobs/GroupLeaderMedicationNotifications.cs
+++ b/Plugins/org.secc.Jobs/GroupLeaderMedicationNotifications.cs
@@ -120,6 +120,7 @@ namespace org.secc.Jobs
             var personEntityType = EntityTypeCache.Get( typeof( Person ) );
             var earliestCheckinDate = RockDateTime.Now.AddDays( -JobAttributes.MedicationCheckinDays ).Date;
             var medicationScheduleAttribute = GetMatrixItemAttribute( "Schedule" );
+            var medicationActiveAttribute = GetMatrixItemAttribute( "MedicationActive" );
 
             foreach ( var groupId in groupIds )
             {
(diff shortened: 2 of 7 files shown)


jwakefield-secc:
## Open question — family-side toggle?

The `MedicationInformation` block (`/myevents` family view) is currently read-only for medication state. Activate/Inactivate is only available in the camp dispense modal (medication manager).

**Question for the team:** Should parents be able to toggle Active/Inactive directly from `/myevents`, or should that stay restricted to camp staff via the dispense page?

**Arguments for keeping it staff-only (current behavior):**
- Cleaner separation: family view = summary, staff view = control
- Prevents a parent from accidentally inactivating a med while a camp is in session (which would silently hide it from that camp's dispense grid)
- The existing `EditWorkflow` block setting already provides a request-based path for parents to propose medication updates

**Arguments for adding to family view:**
- Symmetry — a parent who stops giving a med at home could mark it inactive themselves
- Less waiting on staff for non-camp-time updates

Not blocking this PR — flagging as a follow-up decision.

jwakefield-secc:
## Test verification summary

### ✅ End-to-end verified
- **Phase 1 toggle** on no-snapshot camp (HSM Fall Retreat / page 2818) — toggle flips state, AV row created with `False`, med disappears from dispense grid, reactivates cleanly.
- **Phase 2 Edit/Delete/Add** on snapshot camp (HSM Camp / page 1389, with cloned snapshot data) — all three work; Person master matrix is unchanged after snapshot edits.
- **MedicationInformation block** (`/myevents` family-facing view) — inactive meds correctly hidden from the family medication summary.

### ✅ Verified by SQL predicate test
- **GroupLeaderMedicationNotifications** filter — predicate `IS NULL OR != 'False'` correctly classifies real matrix-item data; TEST ADVIL on test person shows EXCLUDED while Tylenol/Methaline Blue show INCLUDED. The new filter is a `LEFT JOIN + WHERE` inserted into a query that was running successfully in 2025 production cron runs. Full job-trigger test not feasible in dev (no active descendant groups under any reminder-job parent group + concern about real comms to leaders).
- **CreateMedicationLabels** filter — same predicate verified on the same data, same result. The in-memory `.Where + AsBoolean(true)` pattern is identical to the already-end-to-end-verified MedicationInformation filter. No simulator block exists in this env for kiosk-side verification.

### ⏳ Open question (not blocking)
Should `/myevents` allow family-side toggle of Active/Inactive? Currently view-only. See prior comment.

Ready for review.

jwakefield-secc:
## Final verification: stored proc end-to-end

Ran the camp medication report stored procedure directly with real data:

```sql
EXEC dbo._org_secc_CampManager_GetMedicationReport
    @RegistrationTemplateId = 1483,  -- 2025 HIGH SCHOOL CAMP
    @RegistrationInstanceId = NULL;
```

Result for the test person (PersonId 278972, Emilee): **Tylenol** and **Methaline Blue** rows returned. **TEST ADVIL** (which was inactivated via the dispense modal) is correctly absent from the result set.

End-to-end proof for the SP: data flows through the migration 006 `LEFT JOIN AttributeValue medActive` + `WHERE [Value] IS NULL OR [Value] != 'False'` filter, and inactive meds are excluded.

All 7 consumers are now verified. PR is ready to merge.

gmcgrady (commented):
Implementation makes sense, separating snapshot-based camps & person-based camps. Only hesitation is the migration -- I noticed it has some hard coded attribute Ids & an attribute matrix template Guid -- are we sure that these are the same across dev & production? If so, it should be ready to merge.

jwakefield-secc:
> Implementation makes sense, separating snapshot-based camps & person-based camps. Only hesitation is the migration -- I noticed it has some hard coded attribute Ids & an attribute matrix template Guid -- are we sure that these are the same across dev & production? If so, it should be ready to merge.

Yep, I double checked that these hard coded attribute Ids & the attribute matrix template GUID had the same ids. I believe this was the pattern used in the 005 version of this migration as well.

PR #230, opened 2026-05-20
ROCK-8575: Revert Medication Manager to Person-only reads

jwakefield-secc:
## Summary

- Roll back the GroupMember snapshot direction in the Medication Dispense grid. Nothing populates `GroupMember.Medications` in prod, which was causing ~6,300 active campers to render blank in the grid. The dispense grid now reads exclusively from `Person.Medications`, matching every other medication consumer (camp report stored proc, kiosk labels, MyEvents family info block, leader notifications).
- Manage Medications modal opens against the Person master. Edit and Add write to the master, Toggle Active flips `MedicationActive` on the master. **Delete is removed entirely** — staff use the Active toggle to deactivate instead.
- Inactive-but-distributed-today rows stay visible (grayed out, "Inactive" badge, Dispense disabled) so the same-day audit trail isn't lost when a med is deactivated after distribution. Filter is per-(person, medicine, schedule) so a Breakfast distribution doesn't leak Lunch and Dinner rows into the grid for the same med.
- All `MedicationActive` filtering elsewhere (stored proc, kiosk labels, family info block, leader notifications) is unchanged and continues to work correctly against Person.

## What's removed (dead code cleanup)

- `qrySnapshot` LINQ branch, `groupMemberMatrixAttributeIds`, `groupMemberMatrixValuesWithItems`
- `SourcePerson` / `SourceSnapshot` constants
- `IsSnapshotMatrix` method
- `btnDeleteMed_Click` handler
- `hfManageMedsSource` hidden field, `ltManageMedsSourceBadge` literal, `<Rock:DeleteField>` column
- Defense guards in Toggle / Edit / Add handlers
- `source` parameter on the modal configure helper (renamed to `ConfigureManageMedsModal`)

## Performance note

The inactive-but-distributed check uses a `HashSet<string>` keyed by `(PersonId | MatrixItemId | ScheduleGuid)` built once before the per-camper loop. O(1) lookup; safe for heavy-load Camp Freedom days (~40 meds × 200 campers × ~1000 daily distributions).

## Behavioral implications worth flagging to camp ops

- **Edits to medications via the Medication Manager affect the parent's master record globally.** Adding a "Tylenol — as needed" entry for one camp will appear on the camper's master and at every other event they're registered for. Use the Active toggle to deactivate camp-specific additions rather than relying on per-event isolation.
- `GroupMember.Medications` snapshot infrastructure is now formally vestigial in prod. Cleanup of the auto-init Medications attribute on camp GroupTypes is an optional follow-up (separate ticket).

## Files changed

```
.../Reporting/NextGen/MedicationDispense.ascx       (markup)
.../Reporting/NextGen/MedicationDispense.ascx.cs    (code)
```

Net **-14 lines** across the two files.

## Test plan

- [ ] Solution builds cleanly without warnings
- [ ] Medication Dispense grid renders meds for a camper with populated `Person.Medications` (no blank rows)
- [ ] Add a medication via the Manage Medications modal → confirm it appears on the camper's `Person.Medications`
- [ ] Edit an existing medication → confirm changes persist on the Person master
- [ ] Toggle Active/Inactive → confirm `MedicationActive` flips on the Person matrix item
- [ ] Confirm no Delete button is visible anywhere in the modal
- [ ] Distribute a med, then mark inactive on the master → row stays visible, grayed out, "Inactive" badge, Dispense disabled
- [ ] Multi-schedule med: distribute Breakfast, mark inactive → only Breakfast row stays; Lunch and Dinner are hidden
- [ ] "Hide Distributed" toggle still hides all distributed rows including the new inactive-but-dispensed ones
- [ ] Camp Excel report (stored proc) still works unchanged
- [ ] Kiosk medication labels still print correctly at check-in
- [ ] MyEvents family info block still displays correctly

### DIFF
--- a/Plugins/org.secc.Reporting/org_secc/Reporting/NextGen/MedicationDispense.ascx
+++ b/Plugins/org.secc.Reporting/org_secc/Reporting/NextGen/MedicationDispense.ascx
@@ -51,10 +51,8 @@
             <Content>
                 <asp:HiddenField runat="server" ID="hfManageMedsPersonId" />
                 <asp:HiddenField runat="server" ID="hfManageMedsMatrixGuid" />
-                <asp:HiddenField runat="server" ID="hfManageMedsSource" />
                 <div class="row">
                     <div class="col-md-12">
-                        <asp:Literal runat="server" ID="ltManageMedsSourceBadge" />
                         <Rock:Grid runat="server" ID="gMedications" ShowActionRow="false" DataKeyNames="Id" AllowPaging="false"
                             OnRowDataBound="gMedications_RowDataBound">
                             <Columns>
@@ -64,7 +62,6 @@
                                 <Rock:BoolField DataField="Active" HeaderText="Active" />
                                 <Rock:LinkButtonField ID="btnToggleActive" HeaderText="Toggle" CssClass="btn btn-default btn-sm" Text="<i class='fa fa-toggle-on'></i>" OnClick="ToggleActive_Click" />
                                 <Rock:EditField HeaderText="Edit" ID="btnEditMed" OnClick="btnEditMed_Click" />
-                                <Rock:DeleteField HeaderText="Remove" ID="btnDeleteMed" OnClick="btnDeleteMed_Click" />
                             </Columns>
                         </Rock:Grid>
                         <asp:LinkButton runat="server" ID="btnAddMedication" CssClass="btn btn-primary margin-t-sm"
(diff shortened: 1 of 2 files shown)


gmcgrady (approved):
Per conversation with Josh Wakefield, this PR removes "snapshot" functionality from med manager & forces med manager to pull directly from person.Medications (a person attribute). The majority of the changes are removing snapshot-related logic and /or replacing it for person attribute logic. Assuming local testing checks are passing, I am approving & merging for testing in dev.

PR #235, opened 2026-06-12
ROCK-8637 Fix IDOR / broken access control findings

jwakefield-secc:
## Summary

  Closes a set of IDOR / broken-access-control issues where these blocks trusted a client-supplied id (query-string or hidden field) to pick which record to read/write without checking the user was authorized on that specific object now each action re-derives the target server-side and re-authorizes it.

| Row | File | Finding | Resolution |
|-----|------|---------|------------|
| 6 | `GroupManagerBlock.cs` (base) | Base loads `CurrentGroup` from `GroupId` with no authz | Covered by row 10 — base stays unauth by design; all derived blocks self-guard |
| 7 | `FinancialTransactionsExtensionsController` | REST action returns contribution PII | **No action** — POST→`[Secured]` Edit, Staff-only, non-staff get 401; accepted internal risk |
| 8 | `PublicProfileEdit.ascx.cs` | IDOR write — trusts `hfPersonId` on save | **Fixed** — server-side ownership gate against `CurrentPerson.GetFamilies()` |
| 9 | `LinkListEditUsers.ascx.cs` | Privilege escalation — trusts `hfSecurityGroupId` on postback | **Fixed** — re-resolve group + re-run `IsAuthorized(EDIT)` per mutating postback |
| 10 | `GroupDetailLava.ascx.cs` | IDOR read — swap `GroupId` to view arbitrary group PII | **Fixed** — object-level VIEW/EDIT/MANAGE_MEMBERS-or-membership check in `OnLoad` |

## Notes

- Authorization conditions are intentionally **permissive** (VIEW *or* EDIT *or* MANAGE_MEMBERS *or* active membership) so oversight roles (e.g. a coach with EDIT-but-not-VIEW) aren't bounced — they fail closed only for users with no relationship to the group.
- Behavior is unchanged for legitimate users; these add deny-path gates only.

ROCK-8637-PR235-board

### DIFF
--- a/Plugins/org.secc.Cms/org_secc/Cms/LinkListEditUsers.ascx.cs
+++ b/Plugins/org.secc.Cms/org_secc/Cms/LinkListEditUsers.ascx.cs
@@ -262,40 +262,78 @@ namespace RockWeb.Plugins.org_secc.Cms
             }
         }
 
+        /// <summary>
+        /// Re-resolves and re-authorizes the editors security group for a mutating postback.
+        /// The security group must NEVER be trusted from the hfSecurityGroupId hidden field: an
+        /// attacker can tamper it to an arbitrary security role (e.g. RSR - Rock Administration) and
+        /// escalate privileges. Instead re-derive the group from the content channel item and re-run
+        /// the same EDIT authorization check LoadLinkList performs, so the gate runs on every state
+        /// change rather than only on the initial (non-postback) load. Returns null (and shows a
+        /// notification) when the current person is not authorized to edit this list.
+        /// </summary>
+        private int? GetAuthorizedEditorsGroupId()
+        {
+            var contentChannelItemId = hfContentChannelItemId.Value.AsInteger();
+            using (var rockContext = new RockContext())
+            {
+                var contentItem = new ContentChannelItemService( rockContext ).Get( contentChannelItemId );
+                if (contentItem == null
+                    || !contentItem.IsAuthorized( Rock.Security.Authorization.EDIT, CurrentPerson ))
+                {
+                    ShowNotification( NotificationBoxType.Danger,
+                        "<strong><i class=\"fas fa-exclamation-triangle\"></i> Not Authorized</strong><br /> You do not have edit rights to this Link List." );
+                    return null;
+                }
+
+                return GetListEditorsGroup( contentItem ).Id;
+            }
+        }
+
         private void GroupMemberAdd( int personId )
         {
-            var groupId = hfSecurityGroupId.Value.AsInteger();
+            var groupId = GetAuthorizedEditorsGroupId();
+            if (groupId == null)
+            {
+                return;
+            }
 
             using (var rockContext = new RockContext())
             {
                 var defaultRoleId = GroupTypeCache.Get( Rock.SystemGuid.GroupType.GROUPTYPE_SECURITY_ROLE )
                     .DefaultGroupRoleId;
                 var groupService = new GroupService( rockContext );
-                var group = groupService.Get( groupId );
+                var group = groupService.Get( groupId.Value );
                 var groupMemberService = new GroupMemberService( rockContext );
                 groupMemberService.AddOrRestoreGroupMember( group, personId, defaultRoleId.Value );
                 rockContext.SaveChanges();
             }
 
-            LoadEditorList( groupId );
+            LoadEditorList( groupId.Value );
 
         }
 
         private void GroupMemberRemove( int groupMemberId )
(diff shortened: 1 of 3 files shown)


Plugins/org.secc.Cms/org_secc/Cms/LinkListEditUsers.ascx.cs
```diff
@@ -262,40 +262,78 @@ private void LoadLinkList()
...
+        /// notification) when the current person is not authorized to edit this list.
+        /// </summary>
+        private int? GetAuthorizedEditorsGroupId()
+        {
+            var contentChannelItemId = hfContentChannelItemId.Value.AsInteger();
+            using (var rockContext = new RockContext())
+            {
+                var contentItem = new ContentChannelItemService( rockContext ).Get( contentChannelItemId );
+                if (contentItem == null
+                    || !contentItem.IsAuthorized( Rock.Security.Authorization.EDIT, CurrentPerson ))
```
gmcgrady, line 281: This might be changing the paradigm for how security is managed on link lists. Currently, I think (could be wrong) that each link list has an associated group of "allowed editors". The new addition here appears to refer to the 'Edit' security set on the content channel item. I'm not sure that the current allowed editors pattern is using the 'Edit' on the item itself. We'll want to confirm with Stephen Lee. If it does, then we're good to go on this change!

Plugins/org.secc.Cms/org_secc/Cms/PublicProfileEdit.ascx.cs
```diff
@@ -327,6 +327,21 @@ protected void btnSave_Click( object sender, EventArgs e )
...
                 if ( group != null )
                 {
+                    // Ownership gate (IDOR write): ddlGroup and hfPersonId are both client-controlled
+                    // postback values. Validate them against the current person server-side before any
+                    // write: the posted group must be one of the current person's families (ddlGroup is
+                    // populated from CurrentPerson.GetFamilies()), and an existing person being edited
+                    // must be the current person or a member of that family. Reject tampered values.
+                    var allowedFamilyIds = CurrentPerson.GetFamilies().Select( f => f.Id ).ToList();
+                    var postedPersonId = hfPersonId.Value.AsInteger();
+                    if ( !allowedFamilyIds.Contains( group.Id )
```
gmcgrady, line 337: Great way to gate this by forming a list of allowed family ids! I assume front-end user with malicious intent would see the page refresh or partial postback with none of their edits applied. Good to go with this change.

Plugins/org.secc.GroupManager/org_secc/GroupManager/GroupDetailLava.ascx.cs
```diff
@@ -38,6 +38,24 @@ protected override void OnLoad( EventArgs e )
...
+
+            // Object-level authorization (IDOR): the base block loads CurrentGroup straight from the
+            // GroupId request parameter with no authorization. The sibling blocks (GroupRoster,
+            // GroupManagerAttendance, etc.) each enforce their own check; this display block did not,
+            // so an authenticated user could swap GroupId to render an arbitrary group (incl. member
+            // PII) via the Lava template. Follow the same per-block pattern here. The condition is
+            // intentionally permissive (VIEW, EDIT, or MANAGE_MEMBERS, or active membership) so that,
+            // e.g., a coach with EDIT-but-not-VIEW on a group they oversee is not bounced; it only
+            // fails closed for users with no relationship to the group at all.
+            if ( CurrentGroupMember == null
```
gmcgrady, line 50: Elegant way to check any relationship to the group before displaying it. Navigating back to home page is a reasonable state for someone without permissions. Good to go with this change.

gmcgrady (commented):
My only blocking concern is for the pattern used to manage edit access on link lists (see comment on LinkListEditUsers.ascx.cs); if we can verify that the content channel item's edit security is the underlying store for current permissions, then we're good to merge this PR!

EDIT: Channel item security is, in fact, the current underlying store of permissions via a security group per link list.

Plugins/org.secc.Cms/org_secc/Cms/LinkListEditUsers.ascx.cs
gmcgrady: Per conversation with Stephen, the core logic of checking group membership is spot-on here. Also, in regard to my concern: there is a separate security group per link list; however, that group is encompassed by Rock.Security.Authorization.Edit on the item-level check. So, my concern is a moot point. Should be good to merge when ready.

PR #236, opened 2026-06-15
ROCK-8638 Addressing stored XSS vulnerabilities and unsecured REST endpoint

gmcgrady:


Plugins/org.secc.Purchasing/org_secc/Purchasing/PODetail.ascx.cs
```diff
@@ -1198,18 +1198,25 @@ private void LoadVendor( Vendor v, string terms )
             if ( v != null )
```
jwakefield-secc, line 1198: ### PODetail.ascx.cs — 5 remaining RockLiteral XSS sinks need encoding
  
This PR encodes `lblVendorName`, but the original finding listed **six** raw `<Rock:RockLiteral>` sinks. The other five are still assigned raw user-controlled values on this branch:

  | Sink | `.ascx.cs` line |
  |------|-----------------|
  | `lblShipToName.Text = CurrentPurchaseOrder.ShipToName;` | 1096 (also 1929 = `txtShipToName.Text`) |
  | `lblRequisitionItemsTitle.Text = r.Title;` | 2334 |
  | `lblRequisitionItemsRequester.Text = r.Requester.FullName;` | 2335 |
  | `lblIDItemNumber.Text = POI.RequisitionItem.ItemNumber;` | 3343 |
  | `lblIDDescription.Text = POI.RequisitionItem.Description;` | 3345 |

  All six controls are the identical `<Rock:RockLiteral>` type. `RockLiteral` derives from `System.Web.UI.WebControls.Literal` and calls `base.RenderControl` without `Mode = LiteralMode.Encode`, so `.Text` renders as raw HTML —  here's no per-control behavior that makes these five safe. The same reasoning that justified encoding `lblVendorName` applies here.

  **Fix** — wrap each with `System.Web.HttpUtility.HtmlEncode(...)`, e.g.:
  
  ```csharp
  lblShipToName.Text            = System.Web.HttpUtility.HtmlEncode( CurrentPurchaseOrder.ShipToName );
  lblRequisitionItemsTitle.Text = System.Web.HttpUtility.HtmlEncode( r.Title );
  lblRequisitionItemsRequester.Text = System.Web.HttpUtility.HtmlEncode( r.Requester.FullName );
  lblIDItemNumber.Text          = System.Web.HttpUtility.HtmlEncode( POI.RequisitionItem.ItemNumber );
  lblIDDescription.Text         = System.Web.HttpUtility.HtmlEncode( POI.RequisitionItem.Description );

  
  ▎ Note on r.Title: it's encoded at output in RequisitionDetail, but storage stays raw, so it re-surfaces unencoded here.
  
Authenticated-staff-only (not anonymous/public), but the requisition fields are editable by the requester / their ministry and render into a higher-privilege approver's session — a low-priv→high-priv stored-XSS path.
```

**Review by jwakefield-secc — CHANGES_REQUESTED** [review]
A few more RockLiteral controls in `PODetail.ascx.cs` still need encoding (see list below) once those are wrapped, this should be good to go!

File: Plugins/org.secc.Purchasing/org_secc/Purchasing/PODetail.ascx.cs (hunk 1 above)
Comment (gmcgrady) [reply Plugins/org.secc.Purchasing/org_secc/Purchasing/PODetail.ascx.cs:1198]: Resolved by https://github.com/secc/RockPlugins/pull/236/commits/5555fa2a82698d7fc7d62343f6036422192abb60

**Review by jwakefield-secc — COMMENTED** [review]
### PODetail follow-up — 6 sinks ✅, 2 ship-to items to tidy

All six original `RockLiteral` sinks are encoded. Two ship-to fields remain, and they're a special case: `lblShipToName`/`lblShipToAttn` aren't one-way display sinks — the **label is the source of truth** (persisted at save, read back into the edit box, textbox cleared on close), so encoding it in place corrupts the saved value.

---

**1. `ShipToName` double-encodes — our error.** The finding wrongly listed `:1922` (edit-write) as a sink. Since the label is persisted at 1358, encoding both the write (1929) and the read (1096) re-encodes an already-encoded value on each save → `Smith & Co` → `Smith &amp;amp; Co`.

1096 and 1929 aren't encoding the same value twice — they're two separate entry points into the label, and only one runs per value:

- **1096** (load path, `LoadShipTo`) — runs on page load; takes the raw DB value and encodes it in.
- **1929** (edit path, `UpdateShipToAddress`) — runs on modal submit; takes the raw value the user just typed in `txtShipToName` and encodes it in.

A value enters through one door, never both in the same pass, so the label always carries exactly one layer. Both inputs are raw and both must encode, because `RockLiteral` renders `.Text` as raw HTML.

The actual bug spanned a save→reload: 1929 encodes → 1358 persists the *encoded* label to the DB → reload → 1096 encodes the already-encoded value again. The `HtmlDecode` at 1358 breaks the loop by storing raw, so 1096's encode-on-display is the only layer that value ever gets.

Fix — encode **in** to the label (display), decode **out** (lossless, no buildup):

```csharp
1096:  lblShipToName.Text = HttpUtility.HtmlEncode(CurrentPurchaseOrder.ShipToName);  // encode in
1929:  lblShipToName.Text = HttpUtility.HtmlEncode(txtShipToName.Text);               // encode in
1358:  CurrentPurchaseOrder.ShipToName = HttpUtility.HtmlDecode(lblShipToName.Text);  // decode out → raw to DB
1911:  txtShipToName.Text = HttpUtility.HtmlDecode(lblShipToName.Text);               // decode out → raw to edit box
```

---

**2. `lblShipToAttn` display read (1100) still raw** Great find! This is actually the same situation as #1:

```csharp
1100:  lblShipToAttn.Text = HttpUtility.HtmlEncode(CurrentPurchaseOrder.ShipToAttn);  // add encode in
1359:  CurrentPurchaseOrder.ShipToAttn = HttpUtility.HtmlDecode(lblShipToAttn.Text);  // decode out
1912:  txtShipToAttention.Text = HttpUtility.HtmlDecode(lblShipToAttn.Text);          // decode out
```

---

**Why encode/decode instead of storing raw:** the label is the only place a fresh modal edit lives at save time (textbox is cleared on close; `CurrentPurchaseOrder` isn't updated until save). The alternative — a raw-backed hidden field set at all three label-population points (1096/1114/1929) — is more surgery. Encode/decode is the minimal, low-risk fix for a working flow.

Image

gmcgrady:
Addressed the double encoding issue, as found in review, as part of commit https://github.com/secc/RockPlugins/pull/236/commits/3b9e0a473a6d189ce2caed72d3e00fb08b1e3e83

PR #237, opened 2026-06-16
ROCK-8657 Updated action to get person and group from the passed grou…

stphnlee:
…pmember entity


jwakefield-secc (approved):
Looks good to me! I don't think there's any code I'd change and this action is camp specific so it isn't altering any other workflows besides processes that will in fact leverage this change.

gmcgrady (approved):
Clean, well-scoped change. Making the Person and Placement Group attributes optional with a fallback to the triggering GroupMember entity is logically solid, and the attribute-first precedence keeps group-member triggers working unchanged. Approving for merge.

PR #238, opened 2026-06-17
ROCK-8639 Fix IDOR view + PIN privilege-escalation findings

jwakefield-secc:
## Summary
Two High-severity access-control fixes — code-only, safe to merge.

## Changes
- **SuperCheckin** — *PIN privilege escalation.* The guard blocking a PIN for a security-role member showed a warning but fell through (missing `return`) and minted the PIN anyway. Added `return;`.
- **RequisitionDetail** — *horizontal IDOR view.* `RequisitionID` came from the querystring with no ownership check, so any authenticated user could view any requisition. Added `UserCanViewRequisition()` (mirrors `CapitalRequestDetail`) gating `LoadRequisition`/`LoadSummary`/`LoadItems` + the refresh handlers, plus `CreatedBy`/`CurrentPerson` null-safety. Edit-rights ⊆ view-rights, so legitimate requester/creator/ministry/approver/editor flows are unaffected; everyone else fails closed.

## Verification
Two adversarial (read-only) review passes — no blockers/majors. **Compiled and runtime-tested — confirmed working.** Mirrors patterns already on Rock 13.7.

## Not in this PR
- **ManageCommunicationLists** (anonymous account-takeover IDOR) → split to **#239** (coordinated code + WT 250 SMS recipe; would break a live SMS sign-up flow if shipped alone).
- **GroupTracker** (`org.secc.GroupTrackerDemo`) → deferred. Anonymous IDOR write + PII read on `api/GroupTracker/*`, but the plugin is **not deployed** (no RestController/RestAction/EntityType on DEV or PROD; not in the solution; README labels it a demo) and its access model is undefined. **Open questions before any fix:**
  1. What calls it (the Southeast/Subsplash app)? Per-user auth (as the leader), or the `Subsplash API` service account?
  2. Intended audience — leader-only, or leader + staff + admin?
  3. For `GetGroupSummaryByLeader`: does the caller pass the logged-in person's own `personAliasId`, or do staff/admin query others'?
  4. Fold it into the existing, already-secured `GroupApp/*` controllers instead?

### DIFF
--- a/Plugins/org.secc.CheckinMonitor/org_secc/CheckinMonitor/SuperCheckin.ascx.cs
+++ b/Plugins/org.secc.CheckinMonitor/org_secc/CheckinMonitor/SuperCheckin.ascx.cs
@@ -1546,9 +1546,8 @@ try{{
                         {
                             maWarning.Show( "Unable to add PIN to person. This person is in a security role and cannot have a PIN added from this tool.", ModalAlertType.Warning );
                             mdPIN.Hide();
-                            // Must return: without it, execution falls through and a PIN UserLogin is
-                            // created for the security-role member anyway, defeating this guard and
-                            // letting an operator mint a PIN credential for a privileged account.
+                            // Without this return, execution falls through and mints the PIN anyway,
+                            // defeating the security-role guard (privilege escalation).
                             return;
                         }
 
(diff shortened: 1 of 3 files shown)


gmcgrady (approved):
Looks good! Straight forward user view permission gating on the requisition detail. Easy return fix for the PIN escalation. Great choice to roll communication list changes into separate PR. Approving & merging!

One note: the pin escalation fix might inadvertently restrict Kids' volunteers in a way that was previously undesirable but considered "process" -- will be interesting to see if anything in the process was relying on this flawed logic somehow; may not at all, but worth keeping in mind the first weekend it rolls.

PR #241, opened 2026-06-18
ROCK-8675: send Messaging API function key via x-functions-key header

gmcgrady:
Move the decrypted Messaging API function key out of the ?code= query string and into the x-functions-key request header across all MessagingClient methods, so the secret no longer travels in request URLs (captured in Function access logs, App Insights, and proxies). Behavior is otherwise unchanged. Also fixes an Accept-header typo in GetPhoneNumber (applicaiton/json -> application/json).


jwakefield-secc (approved):
Clean! Looks good and based on documentation should work well.

PR #252, opened 2026-06-30
ROCK-8647: embed multi-file media in FormatValue (+AsHtml); clean exports via text path

gmcgrady:
## Summary

Renders multi-file attribute media (`<video>`/``, with a download link beneath each) inline from **both `FormatValue` and `FormatValueAsHtml`**, restoring inline media on Workflow Entry forms and Lava output. Follow-up to #249 (ROCK-8647).

## Why `FormatValue` embeds media (alongside `FormatValueAsHtml`)

Rock does not route all attribute display through `FormatValueAsHtml`. The Workflow Entry form and Lava attribute rendering (`{{ Entity | Attribute:'Key' }}`) resolve the displayed value through `FormatValue`. Confining the embed to `FormatValueAsHtml` (as #249 did) therefore made workflow forms fall back to plain links. Rock's own media field types render media from `FormatValue`: `ImageFieldType` overrides only `FormatValue`, and `MediaElementFieldType`/`MediaWatchFieldType` override both `FormatValue` and `FormatValueAsHtml`. We mirror that here.

To keep the earlier reviewer's concern addressed, raw markup is kept out of exports/plain-text merge via the value-agnostic text channel (see deviation below), and condensed/grid contexts still return the `"N files"` count.

## ⚠️ Deviation from the implementation plan: no `FormatValueAsText` in this Rock v13

The plan specified a `FormatValueAsText` override to keep exports/plain-text merge markup-free. **That override point does not exist in this Rock v13 base `FieldType`** — verified directly against the bundled `Rock.dll` (the `FormatValueAsText` symbol is absent; `FormatValueAsHtml`, `GetTextValue`, and `GetCondensedTextValue` are present). Adding `override FormatValueAsText` would fail to compile (CS0115).

The plan's intent — a clean, markup-free plain-text channel — is implemented via v13's actual successors:

- **`GetTextValue`** → comma-separated filename list (the channel CSV/grid export and plain-text Lava merge use)
- **`GetCondensedTextValue`** → `"N files"`

## Behavior matrix

| Method | Non-condensed | Condensed |
|--------|---------------|-----------|
| `FormatValue` | embed (`<video>`/`` + link) | `"N files"` |
| `FormatValueAsHtml` | embed (delegates to `FormatValue`) | `"N files"` |
| `GetTextValue` / `GetCondensedTextValue` | filename list (no markup) | `"N files"` |

- SVG always renders as a plain link in the embed paths (inline-script XSS guard).
- Files render in `AttributeMatrixItem.Order`.

## Verification

A full MSBuild isn't possible in the plugins-only checkout (the project `ProjectReference`s a Rock source tree not present locally), so the **entire `org.secc.Attributes` assembly was compiled with Roslyn (`langversion:latest`) against the in-repo reference assemblies** — clean build (exit 0). This confirms the `GetTextValue`/`GetCondensedTextValue` overrides bind to real virtual base methods and the assembly type-checks. Runtime checks (workflow form embed, export output, SVG link, ordering) still need a live Rock environment.

### DIFF
--- a/Plugins/org.secc.Attributes/FieldTypes/MultiFileFieldType.cs
+++ b/Plugins/org.secc.Attributes/FieldTypes/MultiFileFieldType.cs
@@ -327,9 +327,15 @@ namespace org.secc.Attributes.FieldTypes
         #region Formatting
 
         /// <summary>
-        /// Link-only formatting. This output also feeds non-HTML consumers — CSV/grid exports
-        /// and Lava merge fields (including emails) — so it must never emit &lt;video&gt;/&lt;img&gt;
-        /// markup. Rich media embedding lives in <see cref="FormatValueAsHtml"/>.
+        /// Embeds media for display: video files render as a native &lt;video&gt; player and images as
+        /// an &lt;img&gt; thumbnail, each with its download link beneath; other types render as a plain
+        /// link. Condensed (grids/summaries) returns an "N files" count.
+        ///
+        /// Rock does not route all attribute display through <see cref="FormatValueAsHtml"/>: the
+        /// Workflow Entry form and Lava attribute rendering ({{ Entity | Attribute:'Key' }}) resolve
+        /// the displayed value through FormatValue. We therefore embed here, mirroring Rock core
+        /// (ImageFieldType embeds via FormatValue; MediaElement/MediaWatch override both). Exports and
+        /// plain-text merge stay markup-free via <see cref="GetTextValue"/> / <see cref="GetCondensedTextValue"/>.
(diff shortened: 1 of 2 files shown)


gmcgrady:
Per discussion with Stephen around Rock RMS's use of `FormatValue` as the display default for the attribute value in form actions, merging this PR to re-implement the native video display as part of the `FormatValue` method. Acknowledge that this may create unintended occurrences of leaky HTML in other output locations, but that could addressed on a case-by-case basis if it proves to be a problem.

PR #256, opened 2026-07-04
ROCK-8702 Add leader import support to Camp Placement Import

stphnlee:
Adds a Campers/Leaders toggle to the column mapping step. Leader imports place people using the group type role marked IsLeader instead of the default role, update existing non-leader members to the leader role, and error when a target group type has no leader role. Preview validates leader roles before processing.

### DIFF
--- a/Plugins/org.secc.Event/README.md
+++ b/Plugins/org.secc.Event/README.md
@@ -2,14 +2,17 @@
 
 > Southeast's custom event/registration blocks — calendar Lava, a heavily-customized registration entry/detail pair, QR event passes, SignNow signing, camp-placement import, and the Community Gives Back program.
 
+Last updated: 2026-07-06
+
 ## Overview
 
 This plugin is a collection of Rock **UI blocks** (`.ascx` user controls) covering Southeast's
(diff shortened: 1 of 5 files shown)


jwakefield-secc (commented):
Reviewed the leader-import logic against the code, Rock 1.13.7 core source, and live DEV import history. The happy path works correctly (verified end-to-end on DEV: add-leader → promote-active-non-leader → skip-existing-leader). One thing worth looking into in Plugins/org.secc.Jobs/Event/CampPlacementImportRunner.cs:

**1. Archived-membership handling is dead code (pre-existing, extended by this PR)**
- GroupMemberService.Queryable() filters out IsArchived = true rows (confirmed from Rock 1.13.7 source), and the plugin uses that plain overload — so no archived row ever enters the lookup.
- The entire if (existingMember.IsArchived) restore/promote branch (and the new leader-role-on-restore logic, plus the missing IsValidGroupMember validation on that path) never executes.
- Consequence: archived members silently get a new active row instead of being restored. The PR built new leader logic on the false assumption that archived rows are visible.
- Fix: either query with archived rows included (Queryable(false, true)) and keep the restore+validate logic, or delete the dead branches.

PR #257, opened 2026-07-06
ROCK-8710: Move parent (Form-key) fields below the waiver text

jwakefield-secc:
## ROCK-8710 — Move parent (Form-key) fields below the waiver text

Layout change to the **Volunteer Signup Form (Connections)** block (`org.secc.Connection` — legacy Web Forms, not Obsidian). The flow becomes: read release → enter parent info → submit.

### What it does
The editable Form-key group-member attribute controls (parent info: ParentFullName / ParentPhone / ParentDOB, etc.) previously rendered inline inside the per-role repeater, above the waiver. They now render **below the Waiver Text**, just above the submit buttons.

- `.ascx`: added `<asp:PlaceHolder ID="phFormAttributes" runat="server" />` **outside** the repeater, positioned after `lWaiverText` and before the `.actions` div.
- `.ascx.cs` `AddGroupMemberAttributes()`: the `Helper.AddEditControls(...)` **write** path now targets `phFormAttributes` (was `phAttributes`). The display-only `Helper.AddDisplayControls(...)` for non-URL attributes still targets the per-role `phAttributes` — unchanged.
- `.ascx.cs` `GetGroupMemberAttributes()`: the matching `Helper.GetEditValues(...)` **read-back** path was also retargeted to `phFormAttributes` (and the now-unused local `phAttributes` in that method removed). Both write and read paths target the same placeholder — if only one had moved, submitted parent-info values would silently stop saving.

### Button label — no code change needed
The original plan's second bullet ("I Agree & Connect") is already satisfied by the existing **"Connect Button Text"** block setting, which is configured to `I Agree & Connect` on the live food-pack block instance (block 6192). An earlier revision of this PR hardcoded the label in code; that was **backed out** because it would have overridden the working per-instance setting for every copy of the block. No button or README changes remain in this PR.

### Known limitation (flagged for reviewer — not fixed here)
`phFormAttributes` is a single **page-level** placeholder, but it's written once per role inside the `foreach (roleRequest in RoleRequests)` loop in `AddGroupMemberAttributes()`. A **multi-role** signup would therefore stack duplicate Form-key controls into the one placeholder (control-tree / ViewState corruption). The current food-pack opportunity is **single-role**, so it's unaffected. Proper multi-role Form-key support would require the placeholder to live inside the repeater item template (with the waiver text moved in alongside it) — a follow-up, not addressed in this ticket.

### DIFF
--- a/Plugins/org.secc.Connection/README.md
+++ b/Plugins/org.secc.Connection/README.md
@@ -62,7 +62,7 @@ Category in Rock: **SECC > Connection**.
 | Setting | Type | Notes |
 |---------|------|-------|
 | Display Phone / Birthdate / Comments | bool | Toggle form fields. |
-| **Connect Button Text** | text (default "Connect") | Submit button label. Note: as of ROCK-8710 the primary submit button label is hardcoded to "I Agree & Connect" in code, so this setting is currently not applied to the primary button. |
+| **Connect Button Text** | text (default "Connect") | Submit button label. |
 | **Lava Template** | code (Lava) | Response message; default `OpportunityResponseMessage.lava`. |
 | Enable Campus Context | bool (default true) | Use page campus context as a filter. |
 | **Connection Status** | defined value | Status for newly created individuals (default Web Prospect). |
--- a/Plugins/org.secc.Connection/README.md
+++ b/Plugins/org.secc.Connection/README.md
@@ -62,7 +62,7 @@ Category in Rock: **SECC > Connection**.
 | Setting | Type | Notes |
 |---------|------|-------|
 | Display Phone / Birthdate / Comments | bool | Toggle form fields. |
-| **Connect Button Text** | text (default "Connect") | Submit button label. |
+| **Connect Button Text** | text (default "Connect") | Submit button label. Note: as of ROCK-8710 the primary submit button label is hardcoded to "I Agree & Connect" in code, so this setting is currently not applied to the primary button. |
 | **Lava Template** | code (Lava) | Response message; default `OpportunityResponseMessage.lava`. |
 | Enable Campus Context | bool (default true) | Use page campus context as a filter. |
 | **Connection Status** | defined value | Status for newly created individuals (default Web Prospect). |
--- a/Plugins/org.secc.Connection/org_secc/Connection/VolunteerSignupFormConnections.ascx
+++ b/Plugins/org.secc.Connection/org_secc/Connection/VolunteerSignupFormConnections.ascx
@@ -66,9 +66,11 @@
 
                 <asp:Literal ID="lWaiverText" runat="server" Visible="false" />
 
+                <asp:PlaceHolder ID="phFormAttributes" runat="server" />
+
                 <div class="actions">
                     <asp:LinkButton ID="btnConnectandAddAnother" Visible="false" runat="server" AccessKey="n" Text="Connect and Add Another" CssClass="btn btn-default" OnClick="btnConnect_Click" />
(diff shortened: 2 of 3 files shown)


stphnlee:
Nice clean change overall — moving the `AddEditControls` write path and `GetEditValues` read path together was the right instinct, and I'm glad the button-label hardcode got backed out in favor of the existing block setting.

One blocking issue before approval (as Copilot noted):

**"Connect and Add Another" will duplicate the Form-key fields.** On that postback, `CreateChildControls` → `AddGroupMemberAttributes()` adds one set of edit controls to `phFormAttributes`, then `btnConnect_Click` calls `ShowDetail(..., true)` followed by `AddGroupMemberAttributes(rockContext)` again, stacking a second set. This used to be harmless because the old target (`phAttributes`) lived inside the repeater and got recreated by the DataBind in `ShowDetail` — but `phFormAttributes` is outside the repeater and never cleared, so the re-shown form renders every parent field twice with duplicate control IDs.

Fix is one line at the top of `AddGroupMemberAttributes()`:

```csharp
phFormAttributes.Controls.Clear();
```

I know "Display Add Another" is off by default and probably off on the food-pack instance, but this block is shared across instances, so let's not ship the regression.

Two non-blocking thoughts, take or leave:

1. On the multi-role limitation you flagged — the failure mode is silent (duplicate control IDs mean role 2+ would inherit role 1's posted values via `GetEditValues`). A cheap guard until the follow-up ticket would be to skip `AddEditControls` when `RepeaterIndex > 0`, or surface a warning when `RoleRequests.Count > 1 && formKeys.Any()`.
2. `GetGroupMemberAttributes` is called per `RepeaterIndex` but now reads the one shared placeholder — a one-line comment there noting it's single-role-only would give the follow-up ticket a breadcrumb at both sites.

Also worth a quick manual test of the Add Another flow with FormKeys populated before merge, since no current instance may exercise that combination.

jwakefield-secc:
Blocker addressed (commit e9c3570).

- `phFormAttributes.Controls.Clear()` added at the top of `AddGroupMemberAttributes()`, running unconditionally on every invocation. This fixes the "Connect and Add Another" duplication: that flow re-invokes `ShowDetail` + `AddGroupMemberAttributes` on the same postback, and since `phFormAttributes` is a page-level placeholder (outside the repeater) it was never being rebuilt/cleared, so a second full set of Form-key controls was stacking in with duplicate IDs (duplicate rendered fields + "Failed to load viewstate" control-tree errors).
- Added an `if ( RepeaterIndex == 0 )` guard around the Form-key `Helper.AddEditControls( ... phFormAttributes ... )` call so it only renders for the first role — `phFormAttributes` is a single shared placeholder. `AddDisplayControls` (per-role `phAttributes`) is unchanged and still runs for every role.
- Added a breadcrumb comment on the read-back side (`GetEditValues( phFormAttributes, ... )` in `GetGroupMemberAttributes`) flagging this as single-role only; multi-role Form-key support (per-role placeholder inside the repeater, waiver moved in) is a follow-up.

Note: I did not build or run this — Rock plugin code can't be compiled in my environment. The Add Another + FormKeys flow still needs a human build-check and manual test (verify no duplicate fields render and Form-key values save on the food-pack instance where `DisplayAddAnother` is enabled) before merge.

jwakefield-secc:
Runtime-tested the Add Another flow on dev (adult + teen, FormKeys populated): a single set of signer fields renders below the waiver on both screens, and both connection requests saved correctly with distinct volunteer vs. signer DOBs with no duplication.

PR #259, opened 2026-07-06
GroupApp: enforce minor communication policy in Communicate endpoint

stphnlee:
## Summary
New org policy: communications may not be sent to minors unless another adult is included.

- Add `IsMinor` to the `GroupAppGroupMember` DTO (set in `GetGroupMembers`, reusing the existing age check)
- `Communicate`: individual emails to minors now CC all parent/guardian emails via `Communication.CCEmails`; sends are rejected with 400 if the minor has no parent email on record
- Minor = `Age.HasValue && AgeClassification != Adult` (unknown age treated as adult, consistent with the roster logic)
- Parents-only (`SendToParents`) and whole-group sends are unchanged

Server-side guard is authoritative; companion GroupApp UI changes ship separately. Deploy this first — it protects even the old app version.

## Test plan
- [ ] Individual send to adult: no CC
- [ ] Individual send to minor with parents: sent, `CCEmails` populated, parents receive one copy
- [ ] Individual send to minor without parent email: 400
- [ ] Individual send with `SendToParents`: unchanged
- [ ] Group send: unchanged, no CC
- [ ] Person with no birthdate: treated as adult

### DIFF
--- a/Plugins/org.secc.Attributes/README.md
+++ b/Plugins/org.secc.Attributes/README.md
@@ -140,5 +140,3 @@ a link. Condensed/grid contexts show an `"N files"` count rather than embedded m
 ---
 
 **Last updated:** 2026-06-29
-</content>
-</invoke>
--- a/Plugins/org.secc.Rest/Controllers/GroupAppGroupMembersController.Partial.cs
+++ b/Plugins/org.secc.Rest/Controllers/GroupAppGroupMembersController.Partial.cs
@@ -133,7 +133,8 @@ namespace org.secc.Rest.Controllers
                         PhotoURL = person.PhotoUrl,
                         IsLeader = groupMember.GroupRole.IsLeader,
                         IsCurrentUser = groupMember.PersonId == currentUser.Person.Id,
-                        IsMinor = isMinor
+                        // Age-derived info is leader-gated like the other PII fields.
+                        IsMinor = isCurrentUserLeader && isMinor
                     };
 
                     // Check if the person is a minor (under 18) and get parent information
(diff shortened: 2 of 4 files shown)


jwakefield-secc (commented):
**🟠 Potential fix — minor test misses kids with no birthdate**

The shared isMinor — person.Age.HasValue && person.AgeClassification != AgeClassification.Adult — requires a birthdate. But Rock classifies a no-DOB child-role family member as AgeClassification.Child with Age = null, so the Age.HasValue half drops them: they're treated as adults and emailed individually with no parent CC and no 400. On DEV that's 16 active children (32 incl. inactive) on GroupApp rosters slipping the guard.

Fix — gate on the classification, which already folds in DOB + family role + IsLockedAsChild:
var isMinor = person.AgeClassification == AgeClassification.Child;
== Child, not != Adult: the only non-Adult values are Child and Unknown, and in the DB every Unknown is a Business (Rock's classifier skips businesses) — no humans are Unknown. So == Child catches every real minor and never misfires; != Adult would treat businesses as minors and 400-block them. Since isMinor also drives the roster's parent-info display, this correctly shows parents for no-DOB minors too — update the README line that says "no birthdate = adult."

---
**⚪ Nit — CC addresses aren't validated**

ccEmails = string.Join(",", parentEmails) dedupes but doesn't validate each address, so one dirty parent email (stray comma/whitespace) can malform the CC list and fail the whole send. Low likelihood — a .Where(e => e.IsValidEmail()) filter would harden it.

gmcgrady (approved):
Solid catch by Josh W to switch to an affirmative `== Child` condition. Overall, this is a clean implementation of the new minor communication policy in the GroupApp's endpoint. See no reason to not merge & test in SEDEV. Approving.

PR #262, opened 2026-07-07
ROCK-8706: Fix Sports & Fitness guest check-in LoadHosts 10-15s hang

jwakefield-secc:
Fixes the 10-15s hang on the Sports & Fitness guest sign-in "load hosts" screen (`GuestCheckin.ascx.cs`, `LoadHosts()`).

### Root cause
`hostsQry` built a `GroupJoin` against the guest-count subquery first, then applied the `OccurrenceId` / `DidAttend` / `EndDateTime` / member-or-employee / `Note` `Where` clauses against the joined projection. EF6 generated a plan that ran the member-ConnectionStatus-OR-employee-subquery predicate over the **full ~9.4M-row `Attendance` table** before narrowing by occurrence — that's the hang.

### Fix
Move those five `Where` clauses onto `attendanceService.Queryable()` directly, applied **before** the `GroupJoin`, so SQL Server narrows by occurrence first. The `GroupJoin`/`DefaultIfEmpty` guest-count join itself is unchanged — a correlated-subquery alternative was tried and rejected (it re-runs the `ValueAsPersonId` scalar UDF per row, measured at 58s).

### Verification (DEV, sample day 2026-01-05, occurrences 648643/45/46/47/50/67/81/85)
- **Row-set equivalence**: old vs. new shape return identical rows — 458/458, 0 mismatches either direction, guest-count sum 1159=1159.
- **Perf**: new shape (narrowed + GroupJoin) = 310ms. Reproducing the pre-fix effective plan (OR predicate + guest join over the full table, no occurrence narrowing) = 16,454ms over 2.87M rows — in the reported hang range. ~53x improvement.
- Adversarial verifier pass (semantic-equivalence + EF6 lens): clean — no blockers/majors/minors. One pre-existing nit noted (missing `AsNoTracking()` at L388, not introduced by this change).

### Scope
Single file, single private method in one block (Block 7417, Sports & Fitness site only). No schema/attribute/index changes.

### DIFF
--- a/Plugins/org.secc.SportsAndFitness/org_secc/SportsAndFitness/ControlCenter/GuestCheckin.ascx.cs
+++ b/Plugins/org.secc.SportsAndFitness/org_secc/SportsAndFitness/ControlCenter/GuestCheckin.ascx.cs
@@ -375,15 +375,27 @@ namespace RockWeb.Plugins.org_secc.SportsAndFitness.ControlCenter
                     .Select(a => a.EntityId);
 
 
-                var hostsQry = attendanceService.Queryable()
-                    .GroupJoin(hostsGuests, a => a.PersonAlias.PersonId, h => h.PersonId,
-                        (a, h) => new {Attendance = a, GuestCount = h.Select(h1 => h1.GuestCount).DefaultIfEmpty()})
-                    .Where( a => occurrenceIds.Contains( a.Attendance.OccurrenceId ) )
-                    .Where( a => a.Attendance.DidAttend == true )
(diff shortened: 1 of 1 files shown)


gmcgrady (approved):
Because all five `.Where()` filters are only dependent on attendance, I agree that they should all be applied prior to the `.GroupJoin()`. Also agree with the subsequent change suggested by co-pilot to keep consistent in our use of `.AsNoTracking()`. This is ready for merge into SEDEV. Approving & merging.

PR #264, opened 2026-07-09
ROCK-8700: Fix double check-in caused by redundant Add after AddOrUpdate

stphnlee:
Rock core's AttendanceService.AddOrUpdate returns the existing tracked attendance when one exists for the same person + occurrence. The extra attendanceService.Add() flipped that entity to Added, so EF inserted a duplicate row and dropped the update closing the old attendance, leaving two active attendances (double check-in).

- SaveAttendance.cs: remove the redundant Add (AddOrUpdate already adds new records); existing records now update in place.
- DataHelper.cs (RoomScanner): comment only. The identical-looking Add in CloneAttendance is intentional clone-by-readd and must not be removed.

*NOTE*: Do not merge in before 7/13

### DIFF
--- a/Plugins/org.secc.FamilyCheckin/Workflows/SaveAttendance.cs
+++ b/Plugins/org.secc.FamilyCheckin/Workflows/SaveAttendance.cs
@@ -182,7 +182,11 @@ namespace org.secc.FamilyCheckin
                                                 }
                                             };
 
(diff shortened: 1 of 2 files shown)


jrwake89 (approved):
Looks good to me, Stephen! I found the same bug when looking into this a bit.

PR #266, opened 2026-07-10
ROCK-8801 Fix Mobile Check In Bug

stphnlee:
Mobile check in was coded to verify the check-in record from the cache, but if the cache key has a null occurrence, then it will throw an error and prevent the check in from completing. This fix has the MCR check the attendance against the database.

Note: this does not change the pre-existing behavior of a room is closing does *not* cancel existing MCRs for that room. This is set up as designed and the children's team knows that if they close a room then they need to move the attendees in that room.

### DIFF
--- a/Plugins/org.secc.FamilyCheckin/README.md
+++ b/Plugins/org.secc.FamilyCheckin/README.md
@@ -104,6 +104,14 @@ Category in Rock: **SECC > Check-in** (medical-consent block: **SECC > Family Ch
 | Kiosk List / Kiosk Detail | Admin CRUD for `Kiosk` records. |
 | Kiosk Type List / Kiosk Type Detail | Admin CRUD for `CheckinKioskType` records. |
 
+**Mobile-reservation validation in QuickCheckin:** when a family with an active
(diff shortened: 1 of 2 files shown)


jwakefield-secc (commented):
Looks solid! Correctly swaps the Null Reference Exception cache walk for an authoritative persisted query, and it's tightly scoped (single caller, read-only, performance difference is negligible). ✅ 

Not blocking; a couple of optional polish items:

README: the Last updated footer is a good add, but doesn't match the format the other plugin READMEs use (Workflow, Attributes, Event, Rest, SmsCapture). Theirs is **Last updated:** <date> — bold label, no --- rule, and file ends with a trailing newline. Current version is plain text, adds a ---, and drops the trailing newline. Could be worth matching for consistency.

Optional / defensive: the guard checks !kioskType.GroupTypeIds.Any() but not == null. In practice SetFromEntity always assigns a list via .ToList(), so this is effectively unreachable — flagging only because ResetGroupLocationSchedules.cs:105 guards GroupTypeIds != null, so adding kioskType.GroupTypeIds == null || to the guard would keep the two consistent.

PR #267, opened 2026-07-13
ROCK-8856 Event Pass: group registrations for the logged-in viewer

stphnlee:
## Summary

Groups event-pass registrations for camp-style events (one registration per camper) into a single carousel, keyed to the **logged-in viewer** rather than the link:

- **Anonymous visitor** — sees exactly the registration/registrant the URL guid names (unchanged from pre-PR behavior). Forwarded links expose nothing beyond themselves.
- **Logged-in viewer** (with the new block setting enabled) — also sees registrants from other registrations in the same registration instance that were made by the viewer or a member of the viewer's family, so a parent who registered each kid separately (or whose spouse registered one) gets all of them in one carousel.

New block setting: **Include Registrar's Registrations** (`IncludeRegistrarRegistrations`, default No). Carousel prev/next arrows are hidden when there is only one pass.

## Fixes over the original approach

The first iteration expanded by the *registrar's family inferred from the URL guid*, which review flagged as an access-widening issue (one registrant guid could unlock other people's pass QR codes — person-level check-in credentials — including non-family registrants and, worst case, everything a staff data-entry registrar had created). Keying expansion to the authenticated viewer removes that class of exposure entirely. Also fixed along the way:

- Waitlist filter now applies before anchoring: with *Include Registrants on Waitlist* = No, a waitlist-only link shows "Pass Not Found" instead of rendering the family's other passes.
- Registrations with a null registrar (`Registration.PersonAliasId` is nullable) no longer crash the page during EF materialization.
- Registrants without a person alias are excluded — they can't render a pass (the template binds `RegistrantPerson.*`) and previously crashed the page.
- A `?Registrant=` deep link now opens with that registrant as the active carousel slide.
- One pass per person when a person appears on multiple registrations (order-preserving `DistinctBy`).
- The per-registrant `PersonSearchKey` lookup (N+1) is replaced with one batched query with a deterministic lowest-Id winner.

## Deployment note

The block setting was renamed from `IncludeFamilyRegistrations` (earlier commits on this branch) to `IncludeRegistrarRegistrations`. Any environment that deployed an earlier branch build will have an orphaned *Include Family Registrations* attribute — delete it and enable the new setting. Fresh deploys from master are unaffected.

## Verification

Query pipeline exercised against a local Rock 13 database via an EF6 integration harness — 17 scenarios covering anonymous vs logged-in, spouse-made registrations included, grandparent/staff-made excluded, waitlist links, null registrar, null alias, no-family viewer, mismatched guid pairs, and dedup. All pass.

### DIFF
--- a/Plugins/org.secc.Event/org_secc/Event/EventPass.ascx.cs
+++ b/Plugins/org.secc.Event/org_secc/Event/EventPass.ascx.cs
@@ -30,6 +30,14 @@ namespace RockWeb.Plugins.org_secc.Event
         Category = "Configuration",
         Key = AttributeKey.IncludeWaitList )]
 
+    [BooleanField( "Include Family Registrations",
+        Description = "Indicates if the pass should also include registrants from other registrations in the same registration instance that were made by members of the registrar's family. Useful for events where each registrant requires a separate registration (e.g. camps). Default is No.",
+        IsRequired = false,
+        DefaultBooleanValue = false,
+        Order = 1,
+        Category = "Configuration",
+        Key = AttributeKey.IncludeFamilyRegistrations )]
+
     [TextField( "Pass Not Found Header",
         Description = "The header/title of the message box that is displayed if the pass is not found.",
         IsRequired = false,
@@ -50,6 +58,7 @@ namespace RockWeb.Plugins.org_secc.Event
         public static class AttributeKey
         {
             public const string IncludeWaitList = "IncludeWaitList";
+            public const string IncludeFamilyRegistrations = "IncludeFamilyRegistrations";
             public const string PassNotFoundHeader = "PassNotFoundHeader";
             public const string PassNotFoundMessage = "PassNotFoundMessage";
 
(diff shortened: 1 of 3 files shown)


Plugins/org.secc.Event/org_secc/Event/EventPass.ascx.cs
```diff
@@ -233,6 +233,8 @@ private void LoadPasses()
             rPassIndicator.DataSource = passes.OrderBy( p => p.ItemOrder );
             rPassIndicator.DataBind();
 
+            phCarouselControls.Visible = passes.Count > 1;
```
gmcgrady, line 236: Slick use of the conditional to gate arrows' visibility.

Plugins/org.secc.Event/org_secc/Event/EventPass.ascx.cs
```diff
@@ -30,6 +30,14 @@ namespace RockWeb.Plugins.org_secc.Event
...
         Key = AttributeKey.IncludeWaitList )]
 
+    [BooleanField( "Include Family Registrations",
+        Description = "Indicates if the pass should also include registrants from other registrations in the same registration instance that were made by members of the registrar's family. Useful for events where each registrant requires a separate registration (e.g. camps). Default is No.",
+        IsRequired = false,
+        DefaultBooleanValue = false,
+        Order = 1,
+        Category = "Configuration",
+        Key = AttributeKey.IncludeFamilyRegistrations )]
+
```
gmcgrady, line 40: Including this as a configurable attribute on the block ensures flexibility. Good choice.

gmcgrady (approved):
Everything looked OK to me--I'd be comfortable approving. Couple notes for consideration from Sonnet 4.6: 
1. EF6 query translation — the expansion predicate uses anchorQry.Any(...) twice inside an IQueryable .Where(). Worth a Database.Log pass to confirm both translate to EXISTS and not a correlated loop or in-memory fallback. (line 158-163)
2. ThenByDescending on a correlated subquery — same concern, one line above .ToList(). If EF6 can't push the bool-valued Any sort key into SQL it may silently pull the full set into memory. (line 181)
---
Both issues stem from the same root: EF6 has to translate C# lambda expressions into SQL, and it doesn't always succeed cleanly.

***For 1*** — the expansion predicate:
anchorQry is itself an IQueryable — it's a composed query that hasn't been executed yet. When you reference it inside another .Where(), EF6 has to translate that into a correlated EXISTS subquery in SQL. Most of the time it does this fine, but when the subquery is complex enough (and anchorQry here involves registration + instance joins), EF6 can fail to translate it and instead:

Pull one or both sets into memory silently and do the join in C#, which could mean loading a large chunk of the registrants table, or throw a runtime exception like "could not be translated into a store expression"

***For 2*** — the ThenByDescending:
Same issue, different context. Using anchorQry.Any(...) as an ORDER BY key is even less common than using it in a WHERE, and EF6's ability to translate a bool-valued correlated subquery into ORDER BY CASE WHEN EXISTS(...) THEN 1 ELSE 0 END is not guaranteed. If it can't, the behavior is unpredictable — it might throw, or it might silently evaluate the sort in-process after pulling more rows than expected.

The reason this might not have surfaced in the offline harness is that test databases are small. Against a production Rock database with thousands of registrants, a mistranslation that causes in-memory evaluation could be slow or blow up entirely.

From Garrett: Approving on the basis that this is likely not an issue since Claude's reasoning is based on the false assumption that our test database is smaller than production.

PR #271, opened 2026-07-17
ROCK-8764: Generate SMS OTP with cryptographic RNG

gmcgrady:
Replace non-cryptographic, clock-seeded System.Random with a cryptographically-secure generator (RandomNumberGenerator) for the SMS one-time login code, using the unbiased float-scaling pattern used by Rock core. Also corrects the pre-existing off-by-one so the full 100000-999999 range is reachable. Output stays a string, so EncodeBcrypt and the SMS merge field are unchanged.

Note: The helper `GetSecureOtp()` method is a re-implementation of the private helper that is used in Rock core. Making it available here for our Authentication plugin. 

Fixes ROCK-8764 (High severity, weak crypto / predictable OTP).


stphnlee (commented):
Looks good! 

Code review with Claude revealed another pre-existing issue that I'm logging in another ticket. 

The only other thing is to update the README (like Copilot said). Claude also made this comment about the README: "Doc comment claims the float-scaling pattern is "unbiased" — it is not, and Rock core's own docs make no such claim.
    "failure_scenario: 2^32 source values map onto 900000 codes: 4294967296/900000 = 4772.19, so 167296 codes draw from 4773 uints and the rest from 4772 (~0.02% bias — security-negligible). A maintainer reading "unbiased" may copy the pattern elsewhere trusting a property it lacks. Fix: soften the comment (e.g. "negligible-bias float-scaling pattern from Rock core")."

gmcgrady:
README updated to reflect changes. All comments from copilot are resolved. Merging.

PR #272, opened 2026-07-17
ROCK-8780: Fix GroupRecommit IDOR — gate person-id params on block EDIT auth

jwakefield-secc:
## ROCK-8780 — GroupRecommit IDOR (High)

**Bug:** `GroupRecommit.LoadPerson()` trusted `PersonId` / `PersonAliasId` from the URL with no authorization check. The block runs as three instances — one internal staff tool plus two on public sites (LWYA, main SECC) that allow AllAuthenticatedUsers — so any anonymous caller, or any logged-in member on the public pages, could swap the id and drive the recommit / group-create flow (creates a `Group`, adds the person as leader, fires a workflow) against an arbitrary person.

## What changed
- **Gate the id params on `IsUserAuthorized( Authorization.EDIT )`** — only staff granted EDIT on the block may resolve an arbitrary person by id. This closes the IDOR.
- **Members are unaffected:** they arrive via the `rckipid` impersonation link or their own login (which set `CurrentPerson` but never pass these params) and fall through the unchanged `CurrentPerson` (self-only) fallback. Anonymous / member id-swaps become no-ops.
- **Null-safe** the `PersonAlias` lookup (was a latent NRE on a bad id).
- **`_person == null` guards** in `btnSave_Click` and `BindGrid` — postback paths that can now run unresolved under the tighter gate; they show the existing login message instead of NRE'ing.

## Required companion (same deploy window)
Auth **Edit-grant on block 884** (internal instance) to the **`RSR - Rock Edit Access`** role. Without it, staff lose the by-id lookup and can only edit themselves. Public blocks (911 / 2945) are intentionally left ungranted — that's what keeps members off the id path on the public sites. Recipe is on the ticket.

## Out of scope
- `GroupMember.GroupTypeId` (v16 NOT NULL) — reverted; no such column in 13.7, deferred to the v16 upgrade.
- Same antipattern in `ChangeEntry` / `FinancialAidFamilyProfile` and the permanent-`rckipid`-token risk — tracked as follow-ups on the ticket.

C# 6 compliant (runtime-compiled `.ascx.cs`). No merge until the companion Auth grant is ready.

### DIFF
--- a/Plugins/org.secc.GroupManager/org_secc/GroupManager/GroupRecommit.ascx.cs
+++ b/Plugins/org.secc.GroupManager/org_secc/GroupManager/GroupRecommit.ascx.cs
@@ -192,6 +192,15 @@ namespace RockWeb.Plugins.org_secc.GroupManager
 
         private void BindGrid()
         {
+            // Grid rebind fires as a postback event even when OnLoad bailed out on a null
+            // person (e.g. an unauthorized id swap that no longer resolves), so guard the
+            // _person / _group dereferences below (ROCK-8780).
+            if ( _person == null || _group == null )
+            {
+                pnlMembers.Visible = false;
+                return;
+            }
+
             var members = _group.Members
                     .Where( gm => gm.PersonId != _person.Id
                     && gm.GroupMemberStatus == GroupMemberStatus.Active )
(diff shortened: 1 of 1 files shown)


jwakefield-secc:
**Visual walkthrough** (request flow, the authorization gate, and how the fix wires into the rest of Rock): https://claude.ai/code/artifact/3a956588-2c85-4580-83c5-2a3e81942435

_Note: this is a private Claude artifact — reviewers need it shared from the artifact's share menu to open it._

gmcgrady (approved):
Straightforward permission check on whether current person can load an arbitrary person. Also, clean null checks throughout avoid throwing unhandled exceptions. Approving & merging, pending testing in SEDEV.

PR #275, opened 2026-07-21
ROCK-8767: remove dormant SystemTest endpoint; harden 3 anonymous plugin REST controllers

jwakefield-secc:
## Summary

Four custom plugin REST controllers were reachable **without authentication** — in Rock, `[Authenticate]` only populates identity, `[Secured]` enforces, and there is no global auth filter, so these had no gate. This removes the dead one and hardens the three that are anonymous **by design** (locking them would break production).

📊 **Visual walkthrough, request flow, and full test plan:** https://claude.ai/code/artifact/7589fb3e-1037-46c2-b299-fb89435daee3

## Changes

- **SystemTest** (`api/systemtest/runtest`) — **endpoint removed.** Dormant (no callers; the monitoring job runs tests directly, not via this controller; 7 of 13 tests ping decommissioned servers). Deleting it closes the surface with zero monitoring impact.
- **QR** (`api/qr/{code}`) — **stays anonymous** (check-in kiosk + emailed/on-page QR images depend on it). Added allowlist `^[A-Za-z0-9-]{1,32}$` (blocks arbitrary/URL content → the phishing/CPU vector) returning 400 on mismatch, plus `using` disposal fixing a GDI/stream leak. Verified **0 of 626,976** real codes (84k AccessKeys + 543k alternate-ids) fail the allowlist — all measure 6–15 characters, so the longest real code is 18 with the Lava filter's `PFP` prefix.
- **Mapping** (`api/mapping/distance`) & **GroupManager** (`api/groupmanager/homegroups`) — **stay anonymous** (public campus/crisis finders). Bounded the billed Azure Maps calls with a destination cap plus a global in-memory **500/min/node compute brake** that degrades to cached-only + alerts when tripped. GroupManager was found by a repo-wide sweep (the ticket named only 3).

## Where the destination cap lives

`GroupUtilities.GetGroupsDestinations` applies `MaxMatrixDestinations = 1000` **after** the GroupLocation + postal-coded Location joins, not on the callers' group queries.

This placement is deliberate. Only groups that survive those joins are ever billed to Azure, so capping the raw group query spends the budget on candidates that produce no destination at all. DEV bears this out: `Camp | Housing` has 2,886 public active groups and **zero** postal-coded locations, while the group types the finders actually use — `Home Group` (263) and `Small Group` (415) — sit far below the cap. A pre-join cap would let unmappable groups crowd out mappable ones; a post-join cap bounds exactly the quantity Azure charges for.

Two supporting details:

- The capped subset is **ordered by group id before `Take`**, so it is stable across requests rather than shuffling.
- The per-group dedupe is expressed as `GroupBy` rather than `DistinctBy`, keeping the chain an `IQueryable` so the cap executes as SQL. This matters because the callers no longer pre-bound their group queries — this method is now the only cap, and it has to bound the database read as well as the Azure call.

## Cost defense on the billed endpoints (layered)

Azure's own ~100-cell/call limit + 50 QPS · existing `LocationDistanceStore` cache (repeat lookups = $0) · the 1000-destination post-join cap · in-memory per-minute compute brake.

**On the brake's real ceiling:** the first request of each minute is always allowed through regardless of size, so that a single legitimate cold-cache request warms the cache instead of stalling forever. The effective per-node ceiling is therefore `max(budget, largest single request)` — with the defaults, up to 1000 computes in a minute rather than a hard 500. It caps the burn *rate* per node, not total spend, and because the counter is in process memory the farm-wide ceiling scales with node count and resets on app pool recycle. A hard dollar ceiling belongs at the Azure Maps account / edge (out of scope — no Azure access).

## Review feedback (second commit)

`ROCK-8767: address review feedback` is stacked separately so it can be diffed on its own:

- Moved the destination cap post-join, added deterministic ordering, and switched the dedupe to `GroupBy` (see above).
- Removed the unreachable `qr == null` branch — QRCoder's `GetGraphic` never returns null, and if it somehow did it threw a 500 from the endpoint this PR is teaching to answer 400.
- Fixed the `shoud` → `should` license header typo in `ComputeBrake.cs` (master corrected this across 580 files in `fee2cdda`; the new file was written against the old spelling).

The branch was also rebased onto master to drop the merge commit, per review. Verified the rebased tree is identical to the pre-rebase tree.

## Two more anonymous-endpoint fixes (third commit)

Found while re-reading the touched controllers:

- `HomeGroupsByDistance` threw an unhandled `ArgumentException` on a malformed postal code, reaching anonymous callers as a **500** — the same failure mode this PR removed from the QR endpoint. Now a **400** via `HttpResponseException`, with a null/whitespace guard for the `NullReferenceException` a null zipcode threw on `.Length`. The method's return type and success-path response shape are unchanged.
- Tightened the QR allowlist from 64 to **32** characters. Code length drives QR version, and `GetGraphic( 20 )` renders 20 px per module, so the ceiling bounds the bitmap this anonymous endpoint allocates per request. Every real code measured 6–15 characters across both sources (84,301 `MobileCheckinRecord.AccessKey` + 542,675 alternate-id `PersonSearchKey`), and the only prefix any caller adds is the Lava filter's 3-char `PFP` — so 18 is the longest real code and 32 keeps ~2× headroom.

## Before merge (not run in CI / not built here)

- **Build** all affected projects. Two things the compiler needs to confirm: QRCoder's `IDisposable` surface, and that the new `GroupBy` / `Min( a => (int?) ... )` projection translates under EF6.
- **Smoke test** (scenarios in the artifact): QR real code → 200 / junk → 400; QR no handle leak under load; SystemTest route → 404; finder pages still return the same groups they did before; brake trips → cached-only (never 500) and recovers next minute.
- **Also:** a >32-char QR code → 400 (and confirm every real kiosk / emailed / Personal Fast Pass code still renders); `api/groupmanager/homegroups/60/abcde` → 400 rather than the 500 it returns today.

## Post-deploy config (per environment)

Optional for function, needed for tuning: create Global Attribute `MappingComputeBudgetPerMinute` (Integer, e.g. 500). The code falls back to 500 if it's absent, so forgetting it doesn't break anything.

## Out of scope / follow-ups

Hard spend cap at the Azure Maps account/edge; evaluate the stale SystemsMonitor plugin for full retirement; the GroupApp controller cluster (session-gated — different, lower-severity gap).

No truncation logging was added when the cap bites. On DEV the largest mappable set (210) sits well under the 1000 cap, so it isn't reachable today, and an unthrottled `ExceptionLogService` write on an anonymous endpoint would reintroduce the same DB-write amplification vector this PR closes. Worth revisiting if a finder is ever pointed at a group type with more than 1000 mapped locations.


gmcgrady (changes requested):
ComputeBrake fallback to cahce-only responses is an elegant way to effectively rate limit in addition to the 1000 groups default from `MaxMatrixDestinations`. Would be comfortable to test in SEDEV; however, I noticed that there's a commit from merging origin/master into this branch... does that commit need to be included on this branch? Can we possibly merge your changes without that merge commit included in this PR?

jwakefield-secc:
Rebased — merge commit gone. Two commits now: the original, plus review fixes kept separate so you can diff just that.

Also moved the `.Take(1000)` post-join into `GetGroupsDestinations`. It was capping the group query before the location joins, so it burned slots on groups with no mapped location — DEV has `Camp | Housing` at 2,886 groups / 0 postal-coded locations, while the actual finder types (`Home Group` 263, `Small Group` 415) are well under the cap. Now ordered before `Take` so the subset is stable, and `GroupBy` instead of `DistinctBy` so the cap runs as SQL.

Correction on my own description: the brake isn't a hard 500/min — first request of each minute always passes, so the real per-node ceiling is up to 1000. Description updated.

PR #276, opened 2026-07-28
ROCK-8760 Remove ticket-in-URL PII leak from SecurityController.Curre…

gmcgrady:
…ntUser

Remove the CurrentUser(string param) overload that decrypted a FormsAuthentication ticket supplied as a URL path segment and returned a person's PII (FullName, LastName, Campus, Gender) to unauthenticated callers, and drop the matching "security" route carrying {param}. Only the session-based CurrentUser() (via the securityNoParam route) remains; it returns the current session user's own data.

Update the org.secc.Rest README to reflect the removed behavior.

Fixes broken access control / sensitive data in URL (CWE-598).

### DIFF
--- a/Plugins/org.secc.Rest/Controllers/SecurityController.Partial.cs
+++ b/Plugins/org.secc.Rest/Controllers/SecurityController.Partial.cs
@@ -37,14 +37,6 @@ namespace org.secc.Rest.Controllers
         public void AddRoutes( HttpRouteCollection routes )
         {
             RouteTable.Routes.MapHttpRoute(
-                name: "security",
-                routeTemplate: "api/org.secc/People/{action}/{param}",
-                defaults: new
-                {
-                    controller = "security",
-                    param = RouteParameter.Optional
-                } ).RouteHandler = new SessionRouteHandler();
-            RouteTable.Routes.MapHttpRoute(
                 name: "securityNoParam",
                 routeTemplate: "api/org.secc/People/{action}",
                 defaults: new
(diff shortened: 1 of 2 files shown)


jwakefield-secc (approved):
## Approve — one doc fix before merge

Fix is correct and complete. Both the `CurrentUser(string param)` overload and the `security` route carrying `{param}` are gone, so the CWE-598 path is removed rather than left live and erroring. `securityNoParam` still resolves `api/org.secc/People/CurrentUser` to the session-based `CurrentUser()` — no behavior change for existing callers.

Good call rejecting the ticket's prescribed `[Authenticate, Secured]` fix: with no `RestAction` record (this controller registers routes directly on `RouteTable.Routes`) it would have 401'd everyone, and `[Authenticate]` alone doesn't block anonymous requests.

**Caller search clean.** Zero references to the `{param}` route in RockPlugins, the `Rock-secc` fork, or the other SECC repos. `LoginJS.ascx:6` hits Rock core's `/api/people/CurrentUser`, unrelated. The endpoint has been effectively dead since ROCK-4992 (`6743003e`, Oct 2020) renamed `api/People/...` → `api/org.secc/People/...` with no caller updates.

### Requested change

- **`README.md:154`** — risk register still lists `CurrentUser(param)` decrypting a URL ticket as an open item. The table at line 56 was updated; this one wasn't. Please update it.

### Optional, same commit

- **`SecurityController.Partial.cs:45`** — `param = RouteParameter.Optional` is now a vestige.
- **`README.md:21`** — "routes" is now one route.

gmcgrady:
Addressed the requested & optional changes from review. Since already approved, merging now.

PR #277, opened 2026-07-28
ROCK-8766: Fix broken access control in ManageSportsPINs

gmcgrady:
AddPIN and RemovePIN derived the target person from hfPersonId, a client-controlled hidden field, allowing any user who reached the block to attach a confirmed PIN UserLogin to any person (e.g. a staff account) and then authenticate as them. Resolve the person server-side from the PageParameter("Person") instead, and require the current user to be authorized to EDIT that person before adding or removing a PIN. Also fix an off-by-one that rejected valid 4-character PINs (pin.Length <= minPinLength -> <).

### DIFF
--- a/Plugins/org.secc.SportsAndFitness/org_secc/SportsAndFitness/ControlCenter/ManageSportsPINs.ascx
+++ b/Plugins/org.secc.SportsAndFitness/org_secc/SportsAndFitness/ControlCenter/ManageSportsPINs.ascx
@@ -20,7 +20,6 @@
                     opacity: 1
                 }
         </style>
-        <asp:HiddenField ID="hfPersonId" runat="server" />
         <asp:Panel ID="pnlMain" runat="server" Visible="false">
             <asp:Panel ID="pnlAlert" runat="server" CssClass="alert alert-warning" Visible="false">
                 <h3><i class="fa fa-exclamation-triangle"></i> <asp:Literal ID="lAlertTitle" runat="server" /></h3>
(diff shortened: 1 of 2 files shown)


jwakefield-secc (approved):
## Approve

Nothing blocks merge. The fix does what ROCK-8766 asked, and I couldn't find a way it breaks anything around it.

Verified the gate isn't a no-op: `Model<T>.ParentAuthority` walks Person instance → Person type default → `GlobalDefault`, which has an explicit Edit / Deny / All Users row. And nothing else is affected — `hfPersonId` was only ever set from `PageParameter("Person")`, so legitimate flows read the same value; no schema or data-shape change, so `SearchResults.SearchByPIN`, `PersonActions.LoadPINList`, SuperCheckin, and Rock check-in are untouched.

### Necessary

Nothing. Two trivial cleanups if you're amending anyway:

1. Revert the file mode change (100644 → 100755): `git update-index --chmod=-x <file>`
2. Delete `hfPersonId` from the `.ascx` and the `.cs`. It's now set and never read. Removing it is what stops this regressing — leaving a populated, client-writable person identifier in the very file this ticket is about invites the next person to reach for it.

### Optional

Gate `LoadPins()`. `ManageSportsPINs.ascx:33` renders `<%# Eval("UserName") %>`, and `UserName` is the PIN — so `?Person=<id>` still displays another person's PINs with no authorization check. One-liner, reusing `GetAuthorizedPerson()`.

I'd call this consistency, not security. Reaching it requires being one of the 17 `CAMPUS - BB Sports and Fitness` members or a Rock admin, and those people already hold Person EDIT over everyone and already manage PINs through the PersonActions modal. Also worth noting: `RequiresRemoteAuthentication` is true, so a PIN isn't a website login — the payoff is impersonation at check-in or the S&F desk.

gmcgrady:
Address necessary changes for cleaning up unused field in https://github.com/secc/RockPlugins/pull/277/commits/cbb8ca0fd135dc95fb0ae55d7e13017133f9dddc. Since already approved, merging now.

PR #278, opened 2026-07-28
ROCK-8763 Prevent open redirect via unvalidated returnurl

gmcgrady:
The OAuth plugin passed the user-supplied returnurl (query string / AuthRedirectCookie) straight into Response.Redirect after only XSS scrubbing, so ?returnurl=https://evil.com could bounce authenticated users to an attacker-controlled site.

- Add shared RedirectUrlHelper.IsLocalUrl that allows only site-relative ("/foo"), app-relative ("~/foo") and same-host absolute URLs, and rejects off-host absolute, protocol-relative ("//evil.com") and backslash ("/\evil.com") variants.
- OAuthServiceProvider: validate returnurl at the source (covers the already-logged-in redirect and the AuthRedirectCookie write) plus a defense-in-depth check at the post-login redirect sink.
- AuthLogout: validate the user-supplied returnurl query parameter; the admin-configured ReturnUrl block setting stays trusted.
- Login: replace the naive private IsLocalUrl (which allowed "//evil.com") with the hardened shared helper.

### DIFF
--- a/Plugins/org.secc.OAuth/Utilities/RedirectUrlHelper.cs
+++ b/Plugins/org.secc.OAuth/Utilities/RedirectUrlHelper.cs
@@ -12,39 +12,71 @@
 // limitations under the License.
 // </copyright>
 using System;
+using System.Linq;
+using Rock.Data;
+using Rock.Model;
+using Rock.Web.Cache;
 
 namespace org.secc.OAuth.Utilities
 {
     /// <summary>
     /// Helper for validating user-supplied redirect ("return") URLs to prevent
-    /// open-redirect attacks (see ROCK-8763). A URL is only treated as safe when
-    /// it navigates back into this application.
+    /// open-redirect attacks (see ROCK-8763).
     /// </summary>
     public static class RedirectUrlHelper
     {
         /// <summary>
-        /// Determines whether the supplied URL is a local URL that is safe to use
-        /// as a redirect target.
+        /// Returns <c>true</c> when the URL is safe to use as a redirect target.
         /// </summary>
         /// <remarks>
-        /// Accepts site-relative paths ("/foo") and application-relative paths
-        /// ("~/foo"), plus absolute URLs whose host matches the current request.
-        /// Rejects protocol-relative URLs ("//evil.com"), backslash variants
-        /// ("/\evil.com"), and absolute URLs pointing at a different host - all of
-        /// which a browser would treat as off-site navigation.
+        /// Accepted:
+        ///   - Site-relative paths ("/foo")
+        ///   - Application-relative paths ("~/foo")
+        ///   - Absolute http/https URLs whose host is on the allow-list
+        ///     (Rock SiteDomain rows, the OAuthAllowedRedirectDomains global attribute,
+        ///     or the current request host).
+        ///
+        /// Rejected:
+        ///   - Strings containing control characters (tabs, newlines, etc.)
+        ///   - Protocol-relative ("//evil.com"), backslash variants ("/\evil.com")
+        ///   - Absolute URLs with non-http/https scheme (e.g. javascript:, data:)
+        ///   - Absolute URLs whose host is not on the allow-list
+        ///
+        /// NOTE: The caller must URL-decode the value before passing it here so that
+        /// encoded control characters (%09, %0A) are expanded before the check runs.
+        ///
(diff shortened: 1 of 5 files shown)


jwakefield-secc (changes requested):
## Review — ROCK-8763

Right approach — validate at the source, one shared helper, admin config stays trusted. Two things to fix before merge.

### 1. This breaks logout across our sites

Today: you log out of `rock.secc.org` → it sends you to `auth.secc.org/logout?returnurl=https://rock.secc.org` → you land back on Rock.

With this PR: that check runs on `auth.secc.org`, asks "is this returnurl the same site as me?", says no, discards it, and falls back to that page's configured default — `https://www.southeastchristian.org`. You get dumped on the public church website.

The site doing the sending is never the destination — that's the whole point of a shared login server. So "same host as the current request" says no to every correct answer. It breaks with two hosts, not twenty.

Verified against prod config: `rock.secc.org` and `moc.secc.org` break. Southeast Christian Church and Group Manager survive only because their configured destination happens to match the fallback.

The ticket asked for "relative **or** on an allow-list." This is missing the allow-list half.

### 2. A tab character still gets through

The check looks at character two and rejects `/` and `\` (so `//evil.com` and `/\evil.com` are caught). Put a tab there instead:

```
?returnurl=/%09/evil.com     →     /<TAB>/evil.com
```

Character two is a tab, so it passes. Rock blocks newlines in redirects but not tabs, so it reaches the `Location` header intact — and browsers strip tabs from the header, turning it into `//evil.com`.

I reasoned this from the code and haven't run it — a web.config setting may neutralize it. Worth a quick real test, since it decides whether the fix actually closes the hole.

### Suggested fix

Reject control characters before the position check:

```csharp
for ( int i = 0; i < url.Length; i++ )
{
    if ( url[i] < ' ' || url[i] == '\u007F' )
        return false;
}
```

Also fixes `?returnurl=/%0Aevil`, which currently passes and then throws an unhandled `ArgumentException` (500 on the login path).

Swap same-host for a host allow-list — read from the `SiteDomain` rows plus one global attribute for `www.southeastchristian.org`, which is Webflow-hosted so Rock has no record of it. Read it live rather than hardcoding, so adding a domain in admin doesn't silently fail to take effect.

⚠️ **Don't use `SiteCache.GetSiteByDomain()` for this:**

```csharp
public SiteDomain GetByDomainContained( string domain )
{
    return Queryable().FirstOrDefault( t => domain.ToLower().Contains( t.Domain.ToLower() ) );
}
```

Substring match — `rock.secc.org.evil.com` contains `rock.secc.org`, so it would call an attacker's domain ours and cache that verdict. Needs exact, case-insensitive equality.

One gotcha when normalizing: Site 32's row is stored as `https://www.southeastchristian.org`, scheme and all, so it'll never match a bare `Uri.Host`.

I have a drafted implementation if it's useful.

### Test before merge

| Input | Expected |
| --- | --- |
| `?returnurl=/%09/evil.com` | blocked ← the one that matters |
| `?returnurl=https://rock.secc.org.evil.com` | blocked (substring trap) |
| `?returnurl=https://evil.com` / `//evil.com` / `/\evil.com` | blocked |
| `?returnurl=https://rock.secc.org` on `auth.secc.org` | allowed |
| Full logout from `rock.secc.org` | returns to Rock, not the church site |

### Notes

- **Deploy together.** These `.ascx.cs` files aren't precompiled — they're copied and compiled at runtime. They now depend on a new type in `org.secc.OAuth.dll`. Ship the DLL and the `.ascx.cs` files in the same drop, or login, logout, and the OAuth SP all fail to load.
- **Separate ticket:** `org.secc.Authentication/.../SMSLogin.ascx.cs:129` has the identical bug on the mobile login path — query `returnurl` → `UrlDecode` → `Response.Redirect`. Out of scope here, but worth filing now. Suggest moving the helper to a shared assembly when that one gets fixed, rather than growing this PR.
- Rest of the plugin is clean — `Authorize.ascx.cs` and the surrounding code only use `Request.RawUrl` or admin settings.
- The `sap9ZMVSR://` deep link on the app-logout block is unaffected — admin setting, never a query param.

gmcgrady (commented):
Implemented an Allowed List of redirect domains as a global attribute configured in Rock. Should resolve earlier review's concern for breaking subdomain redirects.

PR #280, opened 2026-07-28
ROCK-8762 Replace brute-forceable account confirmation code

gmcgrady:
The short confirmation code emailed by api/account/create was MD5( userLogin.Guid ) truncated to six digits, and confirmaccount resolved it by loading every unconfirmed UserLogin and comparing each one's derived code. The code was deterministic, never expired, and a single guess was tested against the whole unconfirmed population, so iterating a 10^6 space could confirm arbitrary accounts.

Randomising the code is not on its own sufficient. The chance a blind guess matches *something* is (live codes / 10^6) no matter how the codes are generated, so this bounds the live-code population and the guess rate as well as making the codes unpredictable.

Verification (api/account/confirmaccount):
- Resolve a code with one equality lookup on a stored SHA-256 hash, restricted to unconfirmed logins, so it matches at most one account. Removes the Queryable().Where( !IsConfirmed ).ToList() scan.
- Expire codes after 60 minutes, consume them on use, and delete the stored rows so dead hashes cannot collide with newly issued codes.
- Budget attempts per forwarded IP and per socket peer, registering before validating so the check and increment are atomic. Nothing is refunded on success: an attacker can mint a valid code for a throwaway account at will, so a refund would be a budget reset they control.
- X-Forwarded-For is read rightmost-first; the leftmost hop is caller supplied and forgeable.
- Client-wide volume is alerted on but never blocked, since every user shares one OAuth client and blocking there would let one attacker stop everyone from confirming.

Issuance (api/account/create):
- Cryptographically random code; only its hash and issue time are stored, in two new UserLogin attributes (org.secc.Migrations 034).
- Cap logins holding a live code per person, so create cannot be looped to stack up simultaneously-valid codes on one person. Scoped to the code lifetime so existing abandoned logins lock nobody out of signup.
- Rate limit create itself, tight per forwarded IP and generous per socket peer. Create is the endpoint that mints codes, and seeding many live codes is what makes guessing cheap.
- Verify the migration has run before writing anything, rather than emailing codes that were never stored.

Attempt budgets are O(1) per key with a bounded key table, so a caller rotating headers cannot grow the tracking state or the work per request.

Rock's native encrypted confirmation token path is untouched.

Residual risk: a six-digit code is retained because users type it by hand. Guess cost still scales with the number of attacking source addresses; see the PR description for the numbers and for the two follow-ups that would close it (expired-code sweep job, and accepting a username alongside the code so the space is per-account).

### DIFF
--- a/Plugins/org.secc.Migrations/034_UserLogin_ConfirmationCodeAttributes.cs
+++ b/Plugins/org.secc.Migrations/034_UserLogin_ConfirmationCodeAttributes.cs
@@ -1,0 +1,68 @@
+// <copyright>
+// Copyright Southeast Christian Church
+//
+// Licensed under the  Southeast Christian Church License (the "License");
+// you may not use this file except in compliance with the License.
+// A copy of the License shoud be included with this file.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using Rock.Plugin;
+
+namespace org.secc.Migrations
+{
+    /// <summary>
+    /// ROCK-8762: backing storage for the short account-confirmation code that
+    /// org.secc.Rest AccountController emails to mobile users.
+    ///
+    /// Only the hash of the outstanding code and the time it was issued are stored. The code
+    /// itself is never persisted, and both values are cleared once the code is used or expires.
+    /// Adding these attributes is metadata only -- no existing rows are read or rewritten.
+    ///
+    /// Attribute keys and the hash attribute's Guid are duplicated as constants in
+    /// AccountController.Partial.cs; keep the two in sync.
+    /// </summary>
+    [MigrationNumber( 34, "1.12.9" )]
+    public partial class UserLogin_ConfirmationCodeAttributes : Migration
+    {
+        public override void Up()
+        {
+            // Entity: Rock.Model.UserLogin Attribute: Confirmation Code Hash
+            RockMigrationHelper.AddOrUpdateEntityAttribute( "Rock.Model.UserLogin",
+                Rock.SystemGuid.FieldType.TEXT,
+                "",
+                "",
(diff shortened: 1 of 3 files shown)


jwakefield-secc (commented):
## Approve

Nothing blocks merge. The fix does what ROCK-8762 asked, and I couldn't find a way it breaks anything around it.

The old flow was unlimited guesses against a permanent, deterministic 10^6 space matched against every unconfirmed login (40,052 of 176,223 on DEV), so a sweep confirmed everything an attacker had seeded. Random code + hash lookup scoped to ≤1 account + expiry + single-use + throttling removes that.

Verified nothing else is affected: `MobileConfirmationCode` and `GetByConfirmationCode` appear in exactly one tracked file, no forks, no internal callers. Migration 34 is unique. Both `AttributeValue` queries pin `AttributeId`, so no cross-entity-type `EntityId` collision, and `UserLogin` attributes are already a working pattern here (`PINPurpose`, 12,473 rows since 2016). `SMS_` logins self-confirm at OTP issuance (`SMSAuthentication.cs:287`), so the new expiry can't strand that path. `SECURITY_CONFIRM_ACCOUNT` still renders `{{ MobileConfirmationCode }}` — no template edit needed.

### Necessary

Nothing.

### Optional

1. **`CONFIRMATION_CODE_LIFETIME_MINUTES` 60 → 1440.** A 60-minute window sits inside normal behaviour — 8,192 confirmed logins since 2020 were first touched in the 5–60 min band vs 1,697 within 5 min. Recovery exists (sign up again under a different username), so it's annoyance not lockout: confusing "Username already exists", intended username burned permanently, dead row left behind. Costs ~480 live codes per attacker address instead of ~20 — ~0.02 hits/hour at 40 guesses/hour, so small.
2. **Delete expired stored hashes opportunistically.** They persist forever today, growing lookup cost until the 10-attempt allocation loop fails and 500s `create`. Slow fuse, but nothing else will clean them up.
3. **Don't charge the create budget before `ModelState.IsValid`** — tripping the password rules burns one of 5 per 15 min, and counting failed validation buys nothing defensively.

gmcgrady:
Created backlog tasks for optional suggestions 2 & 3 from review. Already approved -- merging now.

PR #282, opened 2026-08-04
ROCK-8879 Fix PCR warning display and encode stored requestor comments

stphnlee:
Fixes [ROCK-8879](https://seccdev.atlassian.net/browse/ROCK-8879) · Freshservice [INC-13831](https://southeastchristianchurch.freshservice.com/a/tickets/13831)

## What was broken

Bob reported that the duplicate-mobile-number warning on a profile change request had
stopped rendering — the approver saw raw markup instead of the clickable person links that
used to be there:

```
<h4>Dynamically Generated Warnings:</h4>The phone number (XXX) XXX-XXXX is on the
following records.<ul><li><a href='/Person/XXXXXX' target='_blank'>Jane Doe</a>...
```

This is a regression from our own XSS fix. ROCK-8638 (`ea9c4067`) added a blanket
`HttpUtility.HtmlEncode` to `ltRequestComments` in `ChangeRequestDetail.ascx.cs`.
`RockLiteral` inherits `Literal` with `Mode = Transform`, so that encode is load-bearing for
security — but `RequestorComment` holds a **mix** of plain requestor text and hand-built
markup concatenated at write time, so encoding the whole field at render time also
neutralized the intentional markup. Two write sites produce that markup:

- `ChangeEntry.ascx.cs` → `ValidateMobilePhoneNumber` — the `<h4>`/`<ul>`/`<li>`/`<a>` warning
- `CMPublicProfileEdit.ascx.cs` — the `<br><br>Comment: ` separator

A straight revert was not an option: `CMPublicProfileEdit` is a public-facing block, so
un-encoding at render would reopen the stored XSS ROCK-8638 closed.

## Approach

**Invert where encoding happens. `RequestorComment` is trusted HTML, encoded at every write
site, never at render.** Render time cannot distinguish system markup from user text, so the
decision moves to the point of composition, where the two are still separable.

## Commits

**`042d3e8d` — the display regression**

- All six requestor-text write sites across `ChangeEntry`, `CMPublicProfileEdit` and
  `CMPublicProfileRemovePerson` now run user input through
  `.Trim().EncodeHtml().ConvertCrLfToHtmlBr()`. Multi-line comments keep their line breaks,
  which they lost even before ROCK-8638. The `.Trim()` stops a whitespace-only comment from
  becoming `<br><br>` and rendering an empty comment block.
- `ValidateMobilePhoneNumber` takes an `encodePersonNames` flag and encodes the person names
  it interpolates into its `<li><a>` markup. **This closes a second-order injection that
  predates ROCK-8638** — person names are free text via the public profile block, and they
  were being persisted into `RequestorComment` as markup.
- `ChangeRequestDetail` renders `RequestorComment` raw again, with a comment recording the
  invariant. `ApproverComment` stays encoded at render since it is plain textarea input, and
  now converts line breaks too.

**`bcc7f4c0` — a second sink ROCK-8638 missed**

`ChangeRequests.ascx` renders `DataField="Name"` with `HtmlEncode="false"` so the block can
prepend a warning icon, but the name itself was never encoded. `ChangeRequest.Name` is
assigned in `PreSaveChanges` from the target entity's `ToString()`; for family Group requests
that resolves to `Group.Name`, which is built from user-entered last names. ROCK-8638
encoded this value in the detail header but not in the list block. Now encoded before the
icon is prepended.

**`3f67ba67` — migration 004, the stored rows**

Removing the render-time encode without touching the data would re-arm every row already in
the table. Encoding happened at render under ROCK-8638, so **no write site has ever encoded**
— every existing `RequestorComment` is raw, not just the pre-ROCK-8638 ones. A profile-change
comment containing an `img`/`onerror` payload would execute in an approver's session, and
completed requests stay visible in the list block indefinitely.

`Migrations/004_EncodeStoredRequestorComments.cs` encodes every existing row and converts its
line breaks, mirroring what the write sites now do. Originals are copied to
`_org_secc_ChangeManager_ChangeRequestCommentBackup` first, so `Down()` restores exactly and
the change stays auditable; drop that table once results are spot-checked.

## ⚠️ Deploy ordering

**The migration must ship in the same deploy as the block changes.** It assumes every stored
comment is unencoded, which stops being true the moment the new write sites go live. A Rock
plugin migration is the right vehicle because it runs at app startup before any new comment
can be written, and `PluginMigration` guarantees exactly-once so restarts cannot
double-encode.

Do not run the migration SQL by hand on a database where migration 4 is already recorded —
`Up()` is not idempotent, and a second pass turns `&amp;` into `&amp;amp;`.

## Review notes

Two ordering constraints in the migration's nested `REPLACE` chain are load-bearing, and
getting either wrong produces plausible-looking output:

- `&` must be replaced first, or the entities introduced by the later replacements get
  re-encoded.
- The `<br>` substitution must come last, or its angle brackets get encoded.

I verified the chain against `HttpUtility.HtmlEncode` + `ConvertCrLfToHtmlBr` on 11 inputs —
ampersands, angle brackets, both quote styles, CRLF, bare LF, lone CR, an `img`/`onerror`
payload, whitespace-only, empty, and the ticket's own warning markup. Zero divergence.
Characters above ASCII are deliberately left alone; `HtmlEncode` emits numeric entities for
some, but both forms render identically.

The `encodePersonNames: false` argument at the modal call site is intentional and worth a
close look. `ModalAlert.Show` runs its message through `SanitizeHtml`, which round-trips via
`XmlTextWriter` and re-escapes ampersands — passing pre-encoded HTML there would show
`Smith &amp; Jones` in the modal. The persisted path passes `true`, the modal passes `false`.

I also considered wrapping the render in `SanitizeHtml(false)` for defense in depth and
rejected it: the same `XmlTextWriter` round-trip would reintroduce a display bug of exactly
the shape this PR fixes.

The invariant is now enforced by convention plus code comments across four files, with no
chokepoint. **Any new code writing `RequestorComment` must encode at the write site.** Worth
discussing whether a `SetRequestorComment(string userText)` helper on the model is warranted.

## Testing status

This has been tested against dev database.

## Known follow-up (not in this PR)

The list block's Excel export uses `ExportSource = DataSource`, so an exported entity name
containing `&` now shows `&amp;`. The `<i>` icon markup already leaked into the export before
this change. The proper fix is moving the icon to a `RockTemplateField` and dropping
`HtmlEncode="false"`, which resolves both and removes the need for the manual encode. Noted in
`bcc7f4c0`'s commit message.

🤖 Generated with help from [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QnyxtoVWCf4j2BbMR8w7sq

[ROCK-8879]: https://seccdev.atlassian.net/browse/ROCK-8879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Plugins/org.secc.ChangeManager/Migrations/004_EncodeStoredRequestorComments.cs
+++ b/Plugins/org.secc.ChangeManager/Migrations/004_EncodeStoredRequestorComments.cs
@@ -18,28 +18,37 @@ namespace org.secc.ChangeManager.Migrations
 
     /// <summary>
     /// ROCK-8879. ChangeRequestDetail no longer HTML-encodes RequestorComment at render
-    /// time, because the field holds system-generated markup composed at write time. Every
-    /// row written before that change stored raw, unencoded requestor text, so this
-    /// migration encodes the stored values to match the new write-site behavior.
+    /// time, because the field holds system-generated markup composed at write time. Rows
+    /// written before that change stored raw, unencoded requestor text — except that some
+    /// also contain system-composed markup mixed in (the duplicate-mobile-number warning
+    /// from ChangeEntry and the "Comment:" separator from CMPublicProfileEdit). This
+    /// migration encodes every stored value to neutralize the user text, then restores
+    /// those exact known system fragments so historical warnings keep rendering as markup.
     ///
-    /// This must ship in the same deploy as the block changes: it assumes every existing
-    /// row is unencoded, which stops being true once the new write sites are live. Rock
-    /// runs plugin migrations once, tracked in the PluginMigration table, so it will not
-    /// double-encode on subsequent restarts. Do not re-run the SQL by hand.
+    /// This must ship in the same deploy as the block changes: it assumes existing rows
+    /// are unencoded, which stops being true once the new write sites are live. The whole
+    /// Up() runs inside one IF guarded by the backup table's existence, so a re-run
+    /// (tracked-migration bookkeeping failing after the data commit, a restored database,
+    /// or the SQL being run by hand) is a no-op rather than a double-encode.
     ///
     /// Original values are copied to _org_secc_ChangeManager_ChangeRequestCommentBackup
     /// first, so Down() restores exactly and the change stays auditable. That table can be
-    /// dropped once the results have been spot-checked.
+    /// dropped once the results have been spot-checked — but note dropping it also removes
+    /// the re-run guard, so only drop it once the PluginMigration row is confirmed present.
     /// </summary>
     [MigrationNumber( 4, "1.13.7" )]
     public partial class EncodeStoredRequestorComments : Migration
     {
         public override void Up()
         {
-            // Preserve the originals before touching anything.
+            // Everything runs inside the one IF: the backup table's existence is the
+            // sentinel that the data work already happened, which makes the whole
+            // migration idempotent (Rock commits the migration transaction before it
+            // records the PluginMigration row, so a failure in that gap re-runs Up()).
             Sql( @"
                 IF OBJECT_ID( '_org_secc_ChangeManager_ChangeRequestCommentBackup' ) IS NULL
                 BEGIN
+                    -- Preserve the originals before touching anything.
                     CREATE TABLE [_org_secc_ChangeManager_ChangeRequestCommentBackup] (
                         [ChangeRequestId] INT NOT NULL
                             CONSTRAINT [PK__org_secc_ChangeManager_ChangeRequestCommentBackup] PRIMARY KEY,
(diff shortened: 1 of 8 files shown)


Plugins/org.secc.ChangeManager/Migrations/004_EncodeStoredRequestorComments.cs
```diff
@@ -0,0 +1,110 @@
...
+                                REPLACE(
+                                    REPLACE(
+                                        REPLACE(
+                                            REPLACE( [RequestorComment], '&', '&amp;' ),
+                                        '<', '&lt;' ),
+                                    '>', '&gt;' ),
+                                '""', '&quot;' ),
+                            '''', '&#39;' ),
+                        CHAR(13) + CHAR(10), '<br>' ),
+                    CHAR(10), '<br>' )
```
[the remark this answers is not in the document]
stphnlee: This suggestion is wrong. <br> is the correct tag.

Plugins/org.secc.ChangeManager/org_secc/ChangeManager/ChangeRequestDetail.ascx.cs
```diff
@@ -126,10 +126,20 @@ private void BindGrid()
...
+
+                // RequestorComment is trusted HTML: the blocks that write it compose
+                // system-generated markup (the dynamically generated phone warnings, the
+                // "Comment:" separator) with user text that they encode at that point.
+                // Do not encode here or that markup renders literally (ROCK-8879).
+                // Comments stored before ROCK-8879 were encoded by plugin migration 004
+                // (which also restores the known system-generated fragments), so stored
+                // values are uniformly safe to render without encoding.
+                ltRequestComments.Text = changeRequest.RequestorComment;
             }
```
jwakefield-secc, line 138: Dropping the encode here makes sense, but make sure the render site has protection.

`ltRequestComments.Text = changeRequest.RequestorComment.SanitizeHtml( strict: false );`

jwakefield-secc (approved):
Rendered locally and ensured that it builds and the data displayed correctly. 
Note on DEV: it picked up an interim build of the migration (before the restore step existed), so its 6,287 warning rows were encoded but never restored. I applied the restore and its all pulling through now. Prod is clean (no PluginMigration row, 0 already-encoded rows) and will run the full migration correctly.

Plugins/org.secc.ChangeManager/org_secc/ChangeManager/ChangeRequestDetail.ascx.cs
jwakefield-secc: Missed this comment in the description and makes sense now understanding that Rock's sanitization also encodes:

_I also considered wrapping the render in SanitizeHtml(false) for defense in depth and rejected it: the same XmlTextWriter round-trip would reintroduce a display bug of exactly the shape this PR fixes._

Since we already encode on input at every save path, running it through SanitizeHtml at render would just be a second pass.

PR #291, opened 2026-08-21
Remove stale 2020 prayer-campaign Lava snapshots

stphnlee:
## Summary

Deletes `Content/Lava/18DaysofPrayer/` and `Content/Lava/24HoursofPrayer/` — 44 files, one-time September 2020 snapshots of `VolunteerSignupFormConnections`, `VolunteerSignupWizard`, and `ConnectionOpportunitySearch` (plus their Lava templates and CSS), branched for ROCK-4941 and never synced since.

Each folder's `.ascx.cs` files have exactly two commits in their entire history: the 2020-09-28 creation (`87f64dd5` / `7f3e0984`) and the repo-wide 2026-07-17 `"shoud"` → `"should"` copyright typo sweep. The canonical blocks in `Plugins/org.secc.Connection` have had roughly a dozen functional commits in that window (ROCK-5538, ROCK-6220, the unified phone field in #181, ROCK-8625, ROCK-8710, ROCK-8790), none of which touched the snapshots.

**Why now:** reviewing #290 tripped over this. The snapshots contain their own copy of the `RoleRequests` URL-key merge, so the review flagged them as forked deployed copies needing the same fix. They aren't — but nothing in the tree says so, and the next reviewer would lose the same time.

## Safe to remove

- **Nothing outside the folders references them.** A repo-wide case-insensitive search for `18DaysofPrayer` / `24HoursofPrayer` returns only the self-referential `{% include %}` directives *inside* the two folders. No migration, page-data file, or sibling template points in.
- **Nothing in this repo builds or deploys them.** `org.secc.Connection` has no `.csproj`; `BuildPlugin.ps1` copies only `org_secc`; `Tools/symlink-plugins.ps1` enumerates only `Plugins/`. No `.csproj` anywhere references `Content/Lava`.
- **They could not compile against Rock 1.16 regardless.** They still call the pre-1.8 cache API (`DefinedValueCache.Read`, `CampusCache.Read`, `EntityTypeCache.Read`), `personService.GetByMatch`, and pass `List<string>` where `History.HistoryChangeList` is now required. No working code is being lost.
- **Recoverable** from history at `87f64dd5` / `7f3e0984` if a prayer campaign is ever revived.

## ⚠️ One thing to confirm before merging

`Content/Lava/` is a **hand-maintained mirror of production's RockWeb `~/Content/Lava/` tree** — the `{% include %}` paths in these templates resolve there, and 15 sibling campaign folders (Christmas, Shine, Equip, Missions, …) live alongside. It is not deployed by any script in this repo.

So **deleting these files here does not remove anything from the production web server.** A still-live prayer page keeps working; what's lost is the version-controlled copy.

Could someone who knows the prayer campaigns confirm whether those pages are still live? If they are, the production files should be deleted in the same change so the mirror doesn't silently diverge. If they're long retired, this is pure cleanup.


jwakefield-secc (approved):
Good to go!

PR #293, opened 2026-08-24
ROCK-9069 Add carrier-required SMS disclosures to opt-in surfaces

jwakefield-secc:
## Why

Twilio's carrier compliance review of short code **733733** (ticket #28958126) returned two requirements for the subscription page, where mobile numbers are collected for opt-in:

1. The mandatory SMS disclosures must appear **directly beneath the phone number field** — not elsewhere on the page.
2. Privacy Policy and Mobile Terms must be **properly hyperlinked** (they were plain text).

A page-level notice was added first via block Pre-HTML, but the carrier requires the disclosures adjacent to the input, so this moves them into the block itself.

## What changed

`ManageCommunicationLists.ascx.cs`

- **New `CreateSmsDisclosure()`** — returns a `Literal` with the full disclosure set (program name, message frequency, rates, STOP, HELP) and hyperlinked Privacy Policy / Mobile Terms.
- **Rendered after the phone box on both subscribe paths** — the `"Text Message"` list-type branch and the `CommunicationType.SMS` communication-preference branch.
- **Opt-in confirmation text** now includes the program name and a message frequency disclosure; it previously carried neither.

Rendered output:

> Southeast Christian Church text messages. Message frequency varies. Message & data rates may apply. Reply STOP to opt out, HELP for help. **Privacy Policy** · **Mobile Terms**

Confirmation SMS:

> *[list copy]* Reply Y to confirm msgs from Southeast Christian Church. Msg frequency varies. Reply HELP for help, STOP to quit. Msg&data rates may apply.

## Notes for review

- The disclosure is added as a **sibling** control rather than wrapping the phone box, so the existing `pnlWidget.FindControl($"tbPhone{panelId}")` lookups in `Subscribe()` and the save path continue to resolve.
- Program name and URLs are hardcoded deliberately — these are compliance strings that must stay stable and exact. Worth promoting to block attributes later if we standardize domains (`secc.org` vs `se.church` are both currently in use).
- The confirmation suffix grows by ~43 characters, so some confirmations will more often span two SMS segments. Several already did.
- **The Pre-HTML page-level notice should stay.** The `Subscribe/{keyword}` deep-link flow (`pnlKeyword`) renders a Subscribe button without a phone box, so it is not covered by this change — the page-level notice is what carries disclosures there.

## Testing

- [ ] Expand a Text Message list on `/subscribe` — disclosure renders directly under "Verify Mobile Number"
- [ ] Both links open the correct pages
- [ ] Subscribe still succeeds (confirms the `FindControl` lookups are unaffected)
- [ ] Confirmation SMS arrives with program name and frequency

Related config changes tracked in ROCK-9069 (STOP and START keyword responses).

---

## Review fixes (commits `b950a8ce`..`0b1d4d45`)

Follow-up commits addressing the code-review findings:

**Type resolution unified** (`b950a8ce`) — The keyword panel matched "Text Message" as a substring while the list panel used exact `Equals` against the joined display string, so multi-type lists (including the blank-Type default `"Text Message,Email"`) never rendered the phone box or SMS disclosure. A single `GetListCommunicationSupport`/`GetEffectiveMedium` pair now drives the keyword panel, list-panel render, subscribed-member alerts, and `Subscribe()` — which also fixes `Subscribe()` reading the hardcoded `"Type"` attribute instead of the configurable `AttributeKey` block setting (an NRE when that attribute was missing). **Behavior change:** dual-medium lists previously showed neither subscribed-member alert; they now show the one matching the person's effective medium.

**Consent no longer hinges on retyping** (`e7275474`) — `Subscribe()` only ran `UpdateEmail` when the entered address differed from the person's current email, so a DoNotEmail person subscribing with the prefilled address unchanged stayed DoNotEmail (a silently undeliverable subscription), while editing the address flipped the preference. Every email-list subscribe now runs `UpdateEmail` (EmailAllowed, matching the email box help text) and every SMS-list subscribe runs `UpdateMobilePhone` (re-enables `IsMessagingEnabled`).

**Shared disclosure helper naming 733733** (`b48d6f42`, `cb9547b2`) — The disclosure markup existed as two hand-synced copies (private helper here, inline markup in RequestEventPass) and neither named the short code — and this PR had removed the old help text that was the only user-facing mention of 733733. Both blocks now call a public static `org.secc.Communication.SmsDisclosure.Html()`, and the wording adds the program identification: "*{Org} text messages from 733733.*" The margin parameter is validated against a CSS margin-shorthand pattern so it cannot be a style-injection sink.

**Unit tests** (`57f1ce5b`) — `org.secc.Communication.Tests` covers the helper: org-name encoding, short code and filed links present, malicious margins fall back to the default. 13/13 passing.

### ✅ Compliance sign-off received

The disclosure sentence changed from "*{Org} text messages.*" to "*{Org} text messages from 733733.*" — Twilio gave compliance sign-off on the wording (2026-09-01).

### Notes

- RequestEventPass now depends on `org.secc.Communication.dll` being deployed (same precedent as the Event blocks' existing `org.secc.PersonMatch` dependency).
- Verification: `org.secc.Communication.csproj` builds clean against Rock 1.16.12; the CodeFile block compiles clean via csc with stubbed designer controls; helper output smoke-tested (encoding, short code, margin fallback). Manual page-load scenarios in the Testing checklist above still apply and should be re-run.
- Deferred to follow-up ticket [ROCK-9116](https://seccdev.atlassian.net/browse/ROCK-9116) (out of scope here): `api/account/create` unconditionally sets `IsMessagingEnabled = true` on every new account with no consent field in the payload — a third opt-in surface the same audit likely covers.

[ROCK-9116]: https://seccdev.atlassian.net/browse/ROCK-9116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwakefield-secc:
Standardized the disclosure links on `se.church` (commit f8372b3a).

Verified both domains serve equivalent content — the Text Messaging section's lead sentence is word-for-word identical on each:

| URL | Resolution | SMS section |
|---|---|---|
| `secc.org/terms` | 308 → `my.southeastchristian.org/terms` | yes |
| `se.church/terms` | serves directly (bare → www) | yes |

Went with `se.church` because it is the domain cited in our response to the carrier, so the links a reviewer clicks match the links we gave them.

Worth a separate follow-up: the same terms content exists at two URLs rather than one canonical page with a redirect. That is a drift risk — update one, forget the other — and inconsistent terms is exactly the kind of thing a carrier audit surfaces.

stphnlee:
**Copilot verdicts:**
- Medium (unreachable confirmation): valid, confirmed. `SendConfirmationMessage` has exactly one call site — commented out at line 621. Nothing else calls it, SMS preference branch included. New wording never sends. Worse: even if wired, "Reply Y to confirm" promises double-opt-in gate that not exist — `Subscribe()` sets `GroupMemberStatus.Active` unconditionally, nothing processes Y reply.
- Low (README): valid. `.github/copilot-instructions.md` requires README update + "Last updated" date for behavior changes. `Plugins/org.secc.Communication/README.md` untouched.

**Top findings beyond Copilot:**

1. Keyword deep-link flow uncovered (line 883) — `Subscribe/{keyword}` path gets no in-block disclosure. PR defers to page Pre-HTML notice, but PR's own "Why" says that notice is exactly what failed Twilio requirement # 2. Gap ticket cites stays open.
2. Pre-existing dead-end (line 496) — `CommunicationPreference = RecipientPreference` (DB default 0) hits empty `default:` in touched switch. Subscribe button renders with no input box; click fails "Email Address Reqired" with no email field. User can never subscribe.
3. RequestEventPass gap — other 733733 opt-in surface (`Plugins/org.secc.Event/org_secc/Event/RequestEventPass.ascx:39`) keeps partial disclosure: no program name, no STOP, no linked terms. Carrier compliance attaches to campaign, not one page.
4. Latent NREs in SendConfirmationMessage (line 818) — group `FirstOrDefault` null, empty `SMSSubscribeResponse` leading space, `SystemPhoneNumberCache.Get` null. All bite moment someone un-comments line 621.
5. Hardcoded org name + URLs (line 807) — "Southeast Christian Church" twice, both se.church links compiled in. Repo convention uses OrganizationName global attribute (see SMSAuthentication.cs:41). CMS slug change → dead compliance link until redeploy.
6. Segment cost (line 823) — fixed suffix grew 96→140 GSM-7 chars. Nearly every confirmation now 2+ segments.

Also: boilerplate duplicated with divergent wording, phone-box+disclosure pairing enforced only by comment, WCAG contrast fail (#777 at small size), `target=_blank` no indication, possible phone-input width expansion (needs one visual check), `Subscribe()` reads hardcoded 'Type' key while panel uses configurable AttributeKey setting.

Verified non-issues: FindControl breakage, Controls[4] shift, duplicate IDs, HTML injection, ViewState bloat, & GSM encoding, designer regen, copyright header rule. Both se.church URLs live (200; /terms has Text Messaging section).

Bottom line: PR's on-page disclosures fine, but confirmation-SMS half of PR is claim without wire — either call `SendConfirmationMessage` from Subscribe flow (with null guards, and drop "Reply Y" wording unless gate exists) or strip claim from PR. Update README either way.

jwakefield-secc:
Pushed a04397c9 addressing the review.

**Copilot:**
- **Unreachable confirmation (valid)** — reverted the `SendConfirmationMessage` wording change. Its only call site is commented out at line 621, so the PR was claiming a confirmation-text change that never sends. Wiring the send back up (with null guards, and only with a real Y-reply gate behind the wording) is tracked as follow-up work, not smuggled into a compliance text PR.
- **README (valid)** — block table updated with the disclosure behavior + `Last updated: 2026-08-27`.

**stphnlee:**
1. **Keyword deep-link gap** — fixed. `Subscribe/{keyword}` panel now renders the same disclosures beneath the Subscribe button when the list includes Text Message (`lKeywordSmsDisclosure`).
2. **Dead-end default branch** — fixed. `RecipientPreference` now falls back to the email box; the save path already read the email box for non-SMS preferences, so render and save agree.
3. **RequestEventPass** — fixed. The partial help tooltip is now a visible full disclosure (program name, frequency, rates, STOP/HELP, linked terms + privacy) beneath the delivery-method field, and the Event README is updated.
4. **Latent NREs in `SendConfirmationMessage`** — left as-is deliberately: the method is dormant and this PR no longer touches it. Guards belong in the follow-up that wires the send.
5. **Hardcoded org name** — now `GlobalAttributesCache.Value("OrganizationName")`. The **URLs stay hardcoded on purpose**: they're the exact links filed with the carrier, and a CMS slug change silently breaking a filed compliance link is worse than a redeploy. Documented in the README.
6. **Segment cost** — moot now that the suffix change is reverted.

Also fixed from the minor list: contrast (#595959, was failing WCAG at small size), new-tab links get a `title` indication, and the disclosure HTML is now built in one place (`SmsDisclosureHtml`) so the copies can't drift.

**Not addressed here** (pre-existing, noted for follow-up): `Subscribe()` reading the hardcoded 'Type' key vs the configurable `AttributeKey` setting, and the phone-box/disclosure pairing being enforced by convention rather than structure.

jwakefield-secc:
## Self-review follow-up (82fc6f63, bf6c2653)

Re-read my own review fixes and found three things I'd gotten wrong or left open. Two were real bugs I introduced; all are now resolved.

**1. HTML injection I created — fixed in 82fc6f63.**
Injection was correctly marked a non-issue against the original code, because the org name was a literal. Swapping it for `GlobalAttributesCache.Value( "OrganizationName" )` made it admin-editable text interpolated into markup rendered by a pass-through `Literal` — so my fix opened the hole the review had ruled out. Now `HttpUtility.HtmlEncode`d.

**2. My previous comment was half-true — fixed in 82fc6f63.**
I said the org name now comes from the global attribute. It did in `ManageCommunicationLists`, but I had simultaneously added a *new* hardcoded "Southeast Christian Church" in `RequestEventPass.ascx`. That file now reads the attribute inline (encoded). Verified zero hardcoded instances remain in the branch.

**3. Disclosure placement on `RequestEventPass` — moved in bf6c2653.**
It sat under the delivery-method radio, which is arguably the real consent moment. But the carrier note asked for the disclosures "directly beneath the phone number field," and the auditor is going to screenshot the page. Moved under the Mobile Phone box: matches the ask literally, and the text is now visible regardless of which delivery method is selected.

**Not changed, deliberately (documented in bf6c2653).**
On `/Subscribe/{keyword}` the keyword panel and the full list both render, so expanding a list and switching it to Text Message shows the disclosure twice. Leaving it: each copy sits with its own consent control, and duplicate disclosure is compliant where a missing one is not. Added a comment so it isn't cleaned up as redundancy later.

stphnlee (approved):
LGTM

jwakefield-secc:
## Sahil's follow-up: consent checkbox (8ef6f5c7)

Twilio came back on ticket 28958126 — a disclosure alone isn't enough. They want **an unchecked checkbox carrying a "By checking this box" agreement, placed ahead of the disclosures**, then a screenshot of the full page.

Rebased onto Stephen's `SmsDisclosure` refactor first — his structure is better than what I had, and it collapsed the two SMS render branches and two Subscribe branches into one each, so the gate only needed adding in one place per surface. The consent wording joins the disclosure text in the shared helper (`SmsDisclosure.ConsentText()`), with matching unit tests.

**Made it a real gate, not a decoration.** An unchecked box that still lets you subscribe would fail the re-audit, so:

- **Per-list panels** — `Subscribe` returns early when unticked. Checked *before* the phone write, because `UpdateMobilePhone` sets `IsMessagingEnabled` and that switch is what consent actually gates.
- **Fails closed** — a missing checkbox control means no consent record, so it refuses rather than subscribing. A rendering bug should be loud, not silently non-compliant.
- **Keyword deep link** — consent block moved *above* the Subscribe button so the agreement precedes the action.
- **Request Event Pass** — enforced only on the SMS path; an emailed pass needs no SMS consent.

Already-subscribed people are unaffected — those panels render the unsubscribe toggle, not the opt-in controls, so the checkbox appears only at the real opt-in moment.

### Two things reviewers should weigh in on

**1. Layout is unverified.** `pnlToggle` is a Bootstrap `.btn-group` (inline-block, sizes to widest child), and the phone box + disclosure already live inside it. I bounded the consent block at `max-width:34em` so the sentence can't stretch the group, but **I have no .NET SDK on this machine — nothing here is compiled and the tests are unrun.** Since Sahil wants a full-page screenshot, this needs a real look on DEV before we send anything.

**2. There is still no consent audit record.** The proof of opt-in is "you cannot become a member without ticking the box" plus `GroupMember.DateTimeAdded`. That's defensible and typical for Rock, but if Sahil asks for timestamped proof-of-consent with IP, that's a schema change and a separate ticket — worth deciding now rather than under audit pressure.

PR #305, opened 2026-09-10
ROCK-8991 Send publish group confirmations from noreply@secc.org; leader email becomes Reply-To

jwakefield-secc:
## Summary
- `PublishGroupRegistration.ascx.cs` `SendConfirmation()`: `FromEmail` is now `noreply@secc.org` (same pattern as `GroupAppGroupMembersController` and the home-group editor); the leader-entered `PublishGroup.ConfirmationEmail` is set as `ReplyToEmail` when non-blank. `FromName` unchanged.
- `PublishGroupRequest.ascx`: `tbConfirmationFromEmail` relabeled "Confirmation Reply-To Email" with help text. Control ID unchanged; code-behind untouched (ROCK-9118 branch owns it).
- `org.secc.GroupManager/README.md`: one-line note on send behavior.

No block settings, migrations, or data changes. Existing stored addresses become the Reply-To as-is.

JIRA: https://seccdev.atlassian.net/browse/ROCK-8991

## Review notes
- Real behavior change is for groups whose leader address is `@secc.org` (safe-sender already rewrote external-domain senders to `OrganizationEmail`).

## Test plan
- [x] Build `org.secc.GroupManager` (built and compiled clean during review)
- [ ] Deploy to DEV; open page 2006 for a Small Group and confirm the relabeled field renders and saves
- [ ] Register a test person via `allgroups/registration/{slug}` for a Small Group with a staff-address leader; confirm headers `From: <Leader Name> <noreply@secc.org>` and `Reply-To: <leader address>`
- [ ] Repeat once for a Home Group (Reply-To should equal `noreply@secc.org`)

### DIFF
--- a/Plugins/org.secc.GroupManager/README.md
+++ b/Plugins/org.secc.GroupManager/README.md
@@ -72,7 +72,7 @@ Categories in Rock: **SECC > Groups**, **SECC > Camp**, and **Groups**.
 | Camp Allergies | RockBlock | Campers in a group who have allergies (backed by a stored proc). |
 | Publish Groups Lava | RockBlock | Output PublishGroups via Lava. Settings: `Lava Template`, `Campus Parameter Name` (default `CampusId`), `Category Parameter Name` (default `CategoryId`), `Filter Categories` (defined-value guid list limiting the category filter), `Category Filter Display Mode` (1 = hidden, 2+ = shown). Hides groups whose **Active** member count has reached `GroupCapacity`; group/publish-group attributes load via Rock's bulk `LoadAttributes` (qualifier filtering, inherited attributes, default values). |
 | Publish Group List | RockBlock | List of PublishGroups. |
-| Publish Group Registration | RockBlock | Register a person for a published group. |
+| Publish Group Registration | RockBlock | Register a person for a published group. The confirmation email is sent **From `noreply@secc.org`** with the publish group's `ConfirmationFromName`; the leader-entered `ConfirmationEmail` is used as the **Reply-To**. |
 | Publish Group Request | RockBlock | Display a publish-group request. |
 
 ### Models
--- a/Plugins/org.secc.GroupManager/org_secc/GroupManager/PublishGroupRegistration.ascx.cs
+++ b/Plugins/org.secc.GroupManager/org_secc/GroupManager/PublishGroupRegistration.ascx.cs
@@ -427,8 +427,13 @@ namespace RockWeb.Plugins.org_secc.GroupManager
             mergeObjects["Group"] = _publishGroup.Group;
 
             var message = new RockEmailMessage();
-            message.FromEmail = _publishGroup.ConfirmationEmail;
+            // ROCK-8991: send from a deliverable SECC address; the leader-entered address becomes the Reply-To.
+            message.FromEmail = "noreply@secc.org";
             message.FromName = _publishGroup.ConfirmationFromName;
+            if ( _publishGroup.ConfirmationEmail.IsNotNullOrWhiteSpace() )
+            {
+                message.ReplyToEmail = _publishGroup.ConfirmationEmail;
+            }
             message.Subject = _publishGroup.ConfirmationSubject;
             message.Message = _publishGroup.ConfirmationBody;
             message.AddRecipient( new RockEmailMessageRecipient( person, mergeObjects ) );
--- a/Plugins/org.secc.GroupManager/org_secc/GroupManager/PublishGroupRequest.ascx
+++ b/Plugins/org.secc.GroupManager/org_secc/GroupManager/PublishGroupRequest.ascx
@@ -150,7 +150,8 @@
                             <h3>Confirmation Email</h3>
                             <hr />
                             <Rock:RockTextBox runat="server" ID="tbConfirmationFromName" Label="Confirmation From Name" Required="true" />
-                            <Rock:EmailBox runat="server" ID="tbConfirmationFromEmail" Label="Confirmation From Email" Required="true" />
+                            <Rock:EmailBox runat="server" ID="tbConfirmationFromEmail" Label="Confirmation Reply-To Email" Required="true"
+                                Help="Replies to the confirmation email are sent to this address. The email itself is sent from noreply@secc.org." />
                             <Rock:RockTextBox runat="server" ID="tbConfirmationSubject" Label="Confirmation Email Subject" Required="true" />
                             <Rock:HtmlEditor runat="server" ID="ceConfirmationBody" Label="Confirmation Email Body" Height="400" Required="true" />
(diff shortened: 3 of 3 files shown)


stphnlee:
Reviewed the branch. Four items I'd like addressed before merge. Everything else in the review I'm fine with as-is.

**1. Guard against blank recipient email in `SendConfirmation`**
`PublishGroupRegistration.ascx.cs:419`

The guard only checks `person == null`. Spouse email is optional (`tbSpouseEmail` has no `Required`), and the primary email is optional when the `Require Email` block setting is off. A blank-email recipient reaches the transport and throws in `new MailAddress("")`, which lands in ExceptionLog on every such registration.

```csharp
if ( person == null || person.Email.IsNullOrWhiteSpace() )
{
    return;
}
```

**2. Fall back to `ContactEmail` when `ConfirmationEmail` is blank**
`PublishGroupRegistration.ascx.cs:433-436`

Before this change, a blank `ConfirmationEmail` fell through to `OrganizationEmail` via Rock's `GetFromAddress`, so replies still went somewhere monitored. Now a blank value means no Reply-To at all and replies go to `noreply@`. Legacy rows can have a blank value (the confirmation panel was hidden for some registration modes and `Save` writes whatever is in the textbox). `ContactEmail` is required on the same row and is the natural fallback. The transport already skips a blank Reply-To, so the `if` isn't needed.

```csharp
message.ReplyToEmail = _publishGroup.ConfirmationEmail.IsNotNullOrWhiteSpace()
    ? _publishGroup.ConfirmationEmail.Trim()
    : _publishGroup.ContactEmail;
```

**3. Trim/validate the stored address**
`PublishGroupRequest.ascx.cs:512` and `PublishGroupRegistration.ascx.cs:435`

`Save` stores `tbConfirmationFromEmail.Text` untrimmed and never checks `Page.IsValid`, so `EmailBox` validation is client-side only. Previously a malformed value failed loudly in `new MailAddress(FromEmail)` and nothing was sent. Now it's silently dropped by `SMTPComponent`'s empty catch, or posted verbatim as `h:Reply-To` by Mailgun. Suggest trimming on save and on send (the snippet above already trims), and short-circuiting `Save` when validation fails:

```csharp
// top of Save()
if ( !Page.IsValid )
{
    return;
}

// L512
publishGroup.ConfirmationEmail = tbConfirmationFromEmail.Text.Trim();
```

**4. Home groups: stop forcing `noreply@secc.org` into the Reply-To field**
`PublishGroupRequest.ascx.cs:277-278`

This forced value made sense when the field was the From address. Now it's the Reply-To, so home groups end up with From = Reply-To = `noreply@secc.org`, while the forced body text says "The group leaders will reach out to you shortly." The leader sees a greyed-out "Confirmation Reply-To Email: noreply@secc.org" with help text saying it sets the Reply-To, and reasonably assumes replies reach someone. They don't.

Suggest dropping the forced value and letting the leader supply an address, pre-filled from the contact email when available:

```csharp
// replace L277-278
if ( publishGroup.ContactEmail.IsNotNullOrWhiteSpace() )
{
    tbConfirmationFromEmail.Text = publishGroup.ContactEmail;
}
```

The field stays `Required`, so a blank value can't be saved. This also removes the second `noreply@secc.org` literal in the assembly, and the fallback in item 2 covers any legacy home-group rows.

jwakefield-secc:
Pushed review fixes (rebased onto the help-text/README commit):

1. `SendConfirmation` returns early when `person.Email` is blank.
2. `ReplyToEmail` = trimmed `ConfirmationEmail`, falling back to `ContactEmail` when blank.
3. `Save()` now runs `Page.Validate()` + `Page.IsValid` check (explicit `Validate()` because `Save` is also reached from the `mdlConfirmGroup` modal save click) and trims `ConfirmationEmail`.
4. Home groups: forced `noreply@secc.org` removed; field pre-fills from `ContactEmail`, stays `Required`, remains editable. README row updated to match (the "home groups store noreply@, unmonitored" note is now legacy-only).

Note for merge ordering: item 4 edits the same home-group block in `PublishGroupRequest.ascx.cs` that the unmerged `ROCK-9118-publish-group-request-null-schedule` branch converts to block settings (`HomeGroupConfirmationFromEmail`). Whichever lands second needs a manual merge there, and that block setting becomes redundant after this change.

jwakefield-secc:
Second verification pass turned up problems with my `Page.Validate()` addition; fixed in the latest commit:

- **Removed the bare `Page.Validate()`.** Rock's `ModalDialog` save link already posts back with `CausesValidation = true` (1.16.10 `ModalDialog.cs:478-481`), as do the `BootstrapButton`s, so `Page.IsValid` was already populated on all three `Save()` entry points. The bare call also validated every validation group on the page and would have enforced `iGroupImage.Required = true` (set at the top of `btnPublish_Click`) in the same postback — a live behavior change for non-admin publishes that is not part of this ticket. The `if ( !Page.IsValid ) return;` gate stays.
- **Home-group pre-fill no longer clobbers a stored Reply-To.** Line 248 loads the stored value first; the pre-fill from `ContactEmail` now only runs when the field is blank or still holds the legacy forced `noreply@secc.org`.
- `ContactEmail` fallback trimmed on send, matching `ConfirmationEmail`.

Verified OK on re-review: hidden confirmation panel does not block saves (`HideConfirmationInformation()` sets `Required = false`); null `ReplyToEmail` is safe in the transport; the existing `pRequestor_SelectPerson` handler already pre-fills a blank confirmation email from the contact person, so new home groups are not a hard stop.

**ROCK-9118 merge note (exact conflict):** in the `isHomeGroup` block, 9118 keeps `tbConfirmationFromEmail.ReadOnly = true` and swaps the literal for `GetAttributeValue( HomeGroupConfirmationFromEmail )`; this PR deletes both lines and makes the field editable. A "take both" resolution silently re-locks the Reply-To to `noreply@secc.org` and undoes this ticket with a clean compile. Whichever lands second must drop the `ReadOnly` + forced value; 9118's `HomeGroupConfirmationFromEmail` setting should be removed or repurposed as a pre-fill default.

stphnlee (approved):
Follow-up commits address all four items. Verified:

- `Page.IsValid` is safe on all three `Save` callers (`BootstrapButton` and the `ModalDialog` save link both cause validation, and `RockBlock` stamps the block validation group on every control), and dropping the explicit `Page.Validate()` correctly preserves the old behavior for `iGroupImage`.
- Compile check against `RockWeb\Bin` is clean.

One non-blocking note: existing home-group rows still hold `noreply@secc.org` as `ConfirmationEmail` until the leader re-saves, and the `ContactEmail` fallback only covers blank values. A one-time data update (`ConfirmationEmail = ContactEmail WHERE ConfirmationEmail = 'noreply@secc.org'`) would clean those up if we want replies routed for legacy groups too.

PR #309, opened 2026-09-14
ROCK-9117 Replace ListItems with ValueList in keyword editor; strip blank phrases

jwakefield-secc:
**Problem**
On the Messaging Phone Number Keywords block, adding a "Phrase to Match" row with the "+" button and typing into it saves a blank phrase and silently loses the typed text. Keywords end up with empty entries, and texts like "connect" / "groups" get the wrong response.

**Root cause**
Rock 1.16's `listItems.js` binds its blur/remove handlers only to the inputs present at page load, so a row appended by "+" is never wired up and serializes into the control's hidden field as `""`. The typed text never reaches the server. Verified live on DEV. This PR changes RockPlugins only; no Rock core change.

**Worked example: same keystrokes, Rock 13 vs Rock 16**

Existing keyword has one phrase, `connect`. The editor clicks "+", types `groups` in the new box, clicks Save. Only the control's hidden field is posted; the text boxes have no `name` attribute.

| step | Rock 1.13.7 `listItems.js` | Rock 1.16.x `listItems.js` |
|---|---|---|
| page load | `focusout` and remove-click delegated to `document`, so they cover rows that do not exist yet | `focus`/`blur` and remove-click bound directly to the inputs present right now |
| click "+" | row appended, hidden field rewritten: `["connect",""]` | same: `["connect",""]` |
| type `groups`, blur | delegated `focusout` fires, hidden field rewritten: `["connect","groups"]` | new input has no handler, nothing fires, hidden field still `["connect",""]` |
| click Save | server receives `["connect","groups"]` | server receives `["connect",""]`, `groups` is gone |
| click ✕ on new row | row removed, hidden field rewritten | no handler, nothing happens |

The old block code then copied the hidden field verbatim (`listItems.Select( l => l.Value )`), so the `""` went straight to the Messaging service. Editing an *older* box afterwards, or clicking "+" a second time, re-reads every box and rescues the text, which is why the symptom looked intermittent.

Introduced upstream by SparkDevNetwork/Rock commit `f18e39009` ("Updated List Items control to only post back when item data is changed", 2023-09-13, first shipped in 1.16.0, still present in `develop`). `valueList.js` was never rewritten and still delegates from `document`, hence the swap.

**Change**
- Swap `Rock:ListItems` for `Rock:ValueList` on `MessagingPhoneKeywordList.ascx`. Rock 1.16.6's `valueList.js` delegates `change` and remove-click from `document`, so appended rows serialize correctly.
- Read and write the control with `Rock.Utility.RockSerializableList` (`FromUriEncodedString` / `ToUriEncodedString`), the same helper `ValueList.RenderBaseControl` uses.
- On save: trim every phrase, drop blank/whitespace entries, drop exact duplicates (first-seen order kept). If nothing is left, show the existing validation notification and return before any call to the Messaging service.
- Remove the dead `wordsToMatch` local and the now-unused `Newtonsoft.Json` import in this file.

**How to test on DEV (page 3066)**
1. Edit or add a keyword. Click "+", type a phrase, Save. Reopen and confirm the phrase persisted.
2. Repeat with several rows added in one sitting, and with a row added then removed.
3. Add an empty row and Save with no text: validation message, no blank saved.
4. Enter the identical phrase twice: one stored. Enter `ANGEL` and `Angel`: both survive.
5. Open an existing multi-phrase keyword, change only the Response Message, Save: no phrases lost.
6. Phrase containing a comma and one containing a pipe round-trip intact.

**Harness results (exact 1.16.6 `valueList.js`, server side mirrored as split on `|`, drop empties, unescape, then the block's trim / blank-filter / exact-dedupe)**

| scenario | hidden field posted | phrases the server saves |
|---|---|---|
| stored `connect`, "+" type `groups`, Save | `connect\|groups\|` | `["connect","groups"]` |
| stored `Hello World`, `a,b`, `x\|y`, `café`, `ANGEL`, `Angel`, saved untouched | server-escaped, JS never rewrites | identical, byte for byte |
| same stored list, then one row added (JS rewrites whole field, encodes only `\|` and `,`) | `Hello World\|a%2Cb\|x%7Cy\|café\|…` | all preserved incl. space, comma, pipe, é, both casings |
| typed `  padded  `, an empty row, `connect` again | `…\|  padded  \|\|connect\|` | `padded` once, blank dropped, duplicate dropped |
| only blank rows | `\|   \|` | `[]`, server validation fires, no API call |
| ✕ on a row added this session | row removed, field rewritten | correct |
| typed `50%20off` | `50%20off` | `50 off` (the known limitation above); `100% sure` is left as-is by `Uri.UnescapeDataString` |

**Known limitation**
`valueList.js` encodes only `|` and `,`, so a literal `%xx` sequence typed into a phrase is decoded server-side (`50%20off` becomes `50 off`). Pre-existing Rock `ValueList` behaviour.

The same applies to a *stored* phrase containing `%xx`: it renders correctly in its box, but the moment the editor adds or changes any other row, `valueList.js` rebuilds the whole hidden field from raw box values and the stored `%xx` is decoded on save. Reviewer check on 2026-09-15: 0 of 126 prod phrases match `%[0-9A-Fa-f]{2}`, so nothing is affected today.

**Behaviour change**
`ListItems` allowed drag-reordering phrases; `ValueList` does not. Phrase order is not used anywhere in the plugin (grid joins with `<br />`, search is `Contains`, keyword precedence is the separate `Keyword.Order`), so this is cosmetic.

**Note**
Fixes the editor going forward. Keywords already saved with blank phrases live in the external Messaging service, not the Rock DB, and need re-entry by hand after deploy (separate work item on ROCK-9117).


stphnlee:
Thanks for tackling this — swapping to `ValueList` does sidestep the core `listItems.js` focus/remove regression. I pulled all 126 keywords from the Messaging API and checked them against the findings below; summary table at the bottom. A few requested changes first, then two things to be aware of that don't need action on this PR.

## Requested changes

### 1. Null phrase element crashes the edit form (`.ascx.cs:431`)

`RockSerializableList.ToUriEncodedString` calls `Uri.EscapeDataString(value)` per element with no null guard. The `?? new List<string>()` protects the list, not the elements. `MessagingClient.GetKeyword` is a plain `JsonConvert.DeserializeObject<Keyword>`, so `{"PhrasesToMatch":["connect",null]}` yields a null element → `ArgumentNullException` in `KeywordFormLoad` and the edit form never opens. The old JSON-serialize path tolerated this, as do the other consumers in the block (`string.Join` at ~194, `Contains` at ~474).

No keywords have a null element today, so this is hardening, but it's one line:

```csharp
.Where( p => p != null )   // or .Select( p => p ?? string.Empty )
```

before `ToUriEncodedString`.

### 2. `LoadKeyword()` can return null on save (`.ascx.cs:284`)

Pre-existing, but it's in the function this PR reworked and sits directly below the new early-return. `MessagingClient.GetKeyword` (`MessagingClient.cs:184-201`) only throws on 404; on 500/401/403/502/timeout it falls through to `JsonConvert.DeserializeObject<Keyword>(response.Content)` and returns null, and `keyword.Name = tbName.Text.Trim()` NREs. A `if ( keyword == null ) { NotificationBoxSetContent(...); return; }` would close it while you're here.

### 3. Two "at least one phrase" validators, two messages, two places (`.ascx.cs:274`)

- Zero rows → hidden field `""` → client `HiddenFieldValidator` blocks postback, `ValidationSummary` shows *"Phrases to Match must have at least one value."* (`ValueList.RenderControl` overwrites `RequiredFieldValidator.ErrorMessage`, which is presumably why `RequiredErrorMessage` was dropped from the ascx).
- One blank row → hidden field `"|"` → client passes, full postback, `nbNotifications` shows *"At least one phrase to match is required."*

Same error, two wordings, two locations. Simplest is to drop `Required="true"` on the control and let the server check own it; alternatively reuse `listPhrasesToMatch.RequiredErrorMessage` in the server branch so the wording matches.

### 4. Null-guard placement (`.ascx.cs:268`, `:432`)

- `keyword.PhrasesToMatch ?? new List<string>()` at 432 is effectively unreachable: `LoadPhoneNumberKeywords` copies `item.PhrasesToMatch` into `KeywordSummary` at ~471 (overwriting its `= new List<string>()` default), and `string.Join` at ~193 / `PhraseCount => PhrasesToMatch.Count` crash the grid before any row can be clicked. Better home: `public List<string> PhrasesToMatch { get; set; } = new List<string>();` in `Messaging/Model/Keyword.cs:23`, which covers omitted arrays for all three readers. The API already returns `keywords: null` rather than `[]` for a number with no keywords (the dev test number), so a null `PhrasesToMatch` from a keyword created outside this block isn't far-fetched.
- `listPhrasesToMatch.Value ?? string.Empty` and `( p ?? string.Empty )` at 268 are dead: `HiddenField.Value` never returns null, and `FromUriEncodedString` (`Split` + `UnescapeDataString`) never yields null elements. Rock's own `ValueList.RenderBaseControl` calls it unguarded. Simpler:

```csharp
RockSerializableList.FromUriEncodedString( listPhrasesToMatch.Value ).List
    .Select( p => p.Trim() )
    .Where( p => p.IsNotNullOrWhiteSpace() )
    .Distinct()
    .ToList()
```

## Notes, not blocking

### A. Existing phrases containing a literal `%xx` would be rewritten on unrelated edits

The "known limitation" in the description covers a *freshly typed* `%xx`. The same asymmetry also applies to **stored** data on any save where the editor touched a different row: the server renders `50%20off` as `50%2520off` in the hidden field and `50%20off` in the textbox; clicking `+` or editing any other row makes `valueList.js` rebuild the whole hidden field from raw input values (encoding only `|` and `,`), so `50%20off|newphrase|` is posted, `Uri.UnescapeDataString` turns it into `50 off`, and that's what gets PUT to the Messaging service. No UI cue, and the "saved untouched → identical" test row doesn't exercise this path.

I checked: 0 of 126 phrases in prod match `%[0-9A-Fa-f]{2}`, so nothing breaks on deploy. Worth a line in the PR description / ROCK-9117 so the limitation is documented for the stored-data case too.

### B. Drag-and-drop reorder goes away

`ListItems` wires `.fa-bars` + jQuery sortable; `ValueList` has none. The grid renders phrases in stored order, so with `ValueList` the only way to reorder is delete-and-retype. Not asking for a change, just noting it since the PR body doesn't mention it.

## Prod data check (2026-09-15)

`GET /phonenumbers` from the Messaging API, all numbers, all keywords:

| Check | Count |
|---|---|
| Keywords scanned | 126 |
| Phrases matching `%[0-9A-Fa-f]{2}` (A) | 0 |
| Null phrase element (#1) | 0 |
| Null `PhrasesToMatch` (#4) | 0 |

jwakefield-secc:
Solid checks. All four addressed in 2d548be1:

1. **Null element on edit** — `KeywordFormLoad` filters nulls before `ToUriEncodedString`.
2. **`LoadKeyword` null** — guarded right after the call; shows a Danger notification and returns before any mutation or API call.
3. **Two validators** — kept `Required="true"` so the indicator stays, and changed the server message to Rock's exact client wording ("Phrases to Match must have at least one value."), so it's one sentence either way.
4. **Null guards** — dead coalesces removed. For the DTO default I went one step further than an initializer: Newtonsoft calls the setter on an explicit `"PhrasesToMatch": null`, which would overwrite `= new List<string>()` and take out the grid consumers too. `PhrasesToMatch` now has a backing field with `set { _phrasesToMatch = value ?? new List<string>(); }`, so omitted and explicit null both land as an empty list for every reader.

A and B added to the PR description, with your 0/126 count. README row updated per Copilot.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -260,6 +259,24 @@ private void lbKeywordSave_Click( object sender, EventArgs e )
...
+            // The ValueList hidden field holds a pipe-delimited, Uri-encoded list. Rows the user never typed
+            // into come across as empty entries, so clean them out before deciding whether anything was entered.
+            // Only exact duplicates are dropped: this list is rewritten on every save, including edits that
+            // touch nothing but the response message, so a case-insensitive de-dupe would silently discard
+            // deliberate casing variants that the external matcher may well treat as distinct.
+            var phrasesToMatch = Rock.Utility.RockSerializableList
+                .FromUriEncodedString( listPhrasesToMatch.Value ?? string.Empty ).List
+                .Select( p => ( p ?? string.Empty ).Trim() )
+                .Where( p => p.IsNotNullOrWhiteSpace() )
+                .Distinct()
```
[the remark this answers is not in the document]
jwakefield-secc: Addressed in 2d548be1: the Messaging Phone Number Keywords row now documents the ValueList swap, trim/blank/duplicate handling, the failed-load guard and the loss of drag-reorder, and the last-updated marker is bumped.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -260,11 +259,38 @@ private void lbKeywordSave_Click( object sender, EventArgs e )
...
+            // The ValueList hidden field holds a pipe-delimited, Uri-encoded list. Rows the user never typed
+            // into come across as empty entries, so clean them out before deciding whether anything was entered.
+            // The control's client-side Required validator only catches the zero-row case (a single blank row
+            // serialises as "|" and passes it), so this server check is the real gate. Same wording as the
+            // client message Rock generates for a ValueList, so the user sees one sentence either way.
+            // Only exact duplicates are dropped: this list is rewritten on every save, including edits that
+            // touch nothing but the response message, so a case-insensitive de-dupe would silently discard
+            // deliberate casing variants that the external matcher may well treat as distinct.
+            var phrasesToMatch = Rock.Utility.RockSerializableList
+                .FromUriEncodedString( listPhrasesToMatch.Value ).List
```
stphnlee, line 271: **Security: stored XSS.** `FromUriEncodedString` runs `Uri.UnescapeDataString` on the posted value. A phrase typed as `%3Cimg src=x onerror=alert(document.cookie)%3E` has no literal `<` in the post, so it passes ASP.NET request validation. It is then decoded to real markup and saved through the Messaging API.

`GKeywords_RowDataBound` (line 193) writes the phrases into a `Literal` without HTML-encoding them, so the script runs for everyone who loads the keyword grid. The old `ListItems` JSON path posted a literal `<`, which request validation blocked.

Suggested fix: HTML-encode each phrase in `GKeywords_RowDataBound` (for example `phrases.Select( p => p.EncodeHtml() )`). Optionally, also reject `<`/`>` on save.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -260,11 +259,38 @@ private void lbKeywordSave_Click( object sender, EventArgs e )
...
+            }
+
             bool isNew = false;
             Keyword keyword = null;
             if ( hfKeywordId.Value.IsNotNullOrWhiteSpace() )
             {
                 keyword = LoadKeyword( hfKeywordId.Value );
+                if ( keyword == null )
+                {
+                    // MessagingClient.GetKeyword only throws on 404; any other failure deserializes to null.
```
stphnlee, line 290: The comment's premise only holds for an empty body. Looking at `GetKeyword`:
- A JSON error body (e.g. 500 `{"error":"..."}`) deserializes to a **non-null** `Keyword` with `Id == Guid.Empty`. The guard doesn't fire, and the next Save calls `UpdateKeyword` with blank data.
- A non-JSON body (a 502 HTML page) makes `JsonConvert` throw `JsonReaderException`, so the user gets an unhandled exception instead of this message.

The root fix is in `MessagingClient.GetKeyword`: return null (or throw one known exception) whenever `StatusCode != OK`.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -260,11 +259,38 @@ private void lbKeywordSave_Click( object sender, EventArgs e )
...
+            {
+                NotificationBoxSetContent( "Please correct the following:", "Phrases to Match must have at least one value.", NotificationBoxType.Validation );
+                return;
+            }
+
             bool isNew = false;
             Keyword keyword = null;
             if ( hfKeywordId.Value.IsNotNullOrWhiteSpace() )
             {
                 keyword = LoadKeyword( hfKeywordId.Value );
```
stphnlee, line 287: A 404 still escapes. Say user A opens keyword X and user B deletes it. When A clicks Save, `GetKeyword` throws `Keyword ... not found.` and the async postback shows a Rock exception page, not "Unable to save keyword". Clicking a row on a stale grid (line 426) fails the same way.

The README now says the block shows an error when the API doesn't return a keyword, which isn't true for this case. Suggest catching the 404 here, or having `GetKeyword` return null on 404.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -260,11 +259,38 @@ private void lbKeywordSave_Click( object sender, EventArgs e )
...
+            var phrasesToMatch = Rock.Utility.RockSerializableList
+                .FromUriEncodedString( listPhrasesToMatch.Value ).List
+                .Select( p => p.Trim() )
+                .Where( p => p.IsNotNullOrWhiteSpace() )
+                .Distinct()
+                .ToList();
+
+            if ( !phrasesToMatch.Any() )
+            {
+                NotificationBoxSetContent( "Please correct the following:", "Phrases to Match must have at least one value.", NotificationBoxType.Validation );
```
stphnlee, line 279: `lbKeywordSave_Click` never checks `Page.IsValid`, so the Name and Response Message validators only run in the browser. If client script is bypassed, empty values still reach `AddKeyword`/`UpdateKeyword`.

Also, when the ValueList `HiddenFieldValidator` does fail server-side (empty hidden field), the ValidationSummary shows "Phrases to Match must have at least one value." and this notification repeats the same sentence.

Suggest adding `if ( !Page.IsValid ) return;` near the top, keeping this check for the blank-row (`"|"`) case.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx
```diff
@@ -69,7 +69,7 @@
                     <div class="well">
                         <div class="row">
                             <div class="col-md-4 col-sm-6">
-                                <Rock:ListItems ID="listPhrasesToMatch" runat="server" Label="Phrases to Match" Required="true"  RequiredErrorMessage="At least one phrase to match is required." ValidationGroup="keyword-edit" />
+                                <Rock:ValueList ID="listPhrasesToMatch" runat="server" Label="Phrases to Match" ValuePrompt="Phrase" Required="true" ValidationGroup="keyword-edit" />
```
stphnlee, line 72: Minor: Rock's `valueList.js` binds `$('a.value-list-add').on('click')` directly inside `Sys.Application.add_load`, so every async postback on the page adds another handler to any '+' still in the DOM.

Page 3066 also hosts Messaging Phone Number Detail, which has its own UpdatePanel. After that block does a partial postback with this form open, one '+' click adds two or more rows. Blanks are stripped on save, so no data is lost; it just shows extra rows. `ListItems` had the same class of problem, so this isn't a regression, only a heads-up.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
```diff
@@ -400,6 +424,14 @@ private void KeywordFormLoad( string keywordId )
             if ( keywordId.IsNotNullOrWhiteSpace() )
             {
                 keyword = LoadKeyword( keywordId );
+                if ( keyword == null )
```
stphnlee, line 427: Nit: this null guard, notification and comment are duplicated from `lbKeywordSave_Click` (line 288), and both are patches over `GetKeyword`'s loose status handling. Centralising it in `LoadKeyword`/`GetKeyword` removes both copies and covers future callers.

Separately, `Rock.Utility.RockSerializableList` is fully qualified twice; a `using Rock.Utility;` would be cleaner.

stphnlee (commented):
## Code review (high)

Six findings, most severe first. Checked against Rock 1.16 `ValueList.cs`, `valueList.js`, `RockSerializableList` and `MessagingClient.GetKeyword`.

1. **Security: stored XSS through percent-encoded phrases.** `FromUriEncodedString` URL-decodes the posted value, so `%3Cscript%3E` passes ASP.NET request validation and is stored as a literal `<script>`. `GKeywords_RowDataBound` (line 193, not in this diff) writes phrases into a `Literal` without encoding. Fix: HTML-encode phrases when rendering the grid.
2. **`GetKeyword` failure modes the null guards miss.** A JSON error body deserializes to a non-null `Keyword` with `Id == Guid.Empty`. A non-JSON body (e.g. 502 HTML) throws `JsonReaderException`. Fix at the source: have `GetKeyword` return null (or throw) whenever `StatusCode != OK`. `MessagingClient.cs` isn't in this diff, so that part is only in this summary.
3. **404 still throws.** A keyword deleted mid-edit, or clicked on a stale grid, never reaches the new null checks. That contradicts the README's claim that the block shows an error instead.
4. **No `Page.IsValid` check in `lbKeywordSave_Click`.** Name and Response Message are only validated in the browser. When the ValueList validator does fail server-side, the phrase error appears twice.
5. **`valueList.js` stacks '+' handlers on async postbacks.** Another UpdatePanel on page 3066 doing a partial postback makes one '+' click add several rows. Cosmetic, because blanks are stripped on save.
6. **Cleanup.** The null-guard block and its comment are duplicated in save and load; centralising the status check in `GetKeyword` removes both. `Rock.Utility.RockSerializableList` is fully qualified twice where a `using Rock.Utility;` would do.

The inline comments below cover the diff lines.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
jwakefield-secc: Fixed in 84166eb8. `GKeywords_RowDataBound` now HTML-encodes each phrase before joining with `<br />` (`item.PhrasesToMatch.Select( p => p.EncodeHtml() )`). The other grid columns are bound fields, which already encode. Checked locally: a stored `` renders as `&lt;img src=x onerror=alert(1)&gt;`.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
jwakefield-secc: Fixed at the source in 84166eb8. `MessagingClient.GetKeyword` now returns null on any non-OK status, on a body that fails to deserialize (`JsonException`), and on JSON that yields `Id == Guid.Empty`. The block is its only caller. Checked locally against a stub server: 500 with a JSON error body, 502 with an HTML page, and a 200 with non-keyword JSON all come back null, and a valid 200 still returns the keyword.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
jwakefield-secc: Fixed in 84166eb8. `GetKeyword` no longer throws on 404; it returns null like every other failure, so a keyword deleted mid-edit or clicked on a stale grid now shows "Unable to load keyword … It may have been deleted, or the service may be unavailable." instead of an exception page. README row updated to name the deleted-keyword case.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
jwakefield-secc: Fixed in 84166eb8. `lbKeywordSave_Click` now starts with `if ( !Page.IsValid ) return;`, so Name / Response Message are enforced server-side. When the ValueList validator fails server-side, the handler returns before the block's own phrase message, so the sentence only appears once (in the ValidationSummary). The server check stays for the blank-row (`"|"`) case.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx
jwakefield-secc: Agreed, and confirmed in 1.16 `valueList.js`: the "+" handler is bound directly inside `Sys.Application.add_load`, so each partial postback elsewhere on the page stacks another one. No change here. It is Rock core behaviour, `ListItems` had the same class of problem, the extra rows are blank and stripped on save, and reopening the form resets it. Working around it would mean unbinding core handlers from plugin script, which is not worth it for a cosmetic issue.

Plugins/org.secc.Communication/org_secc/Communication/MessagingPhoneKeywordList.ascx.cs
jwakefield-secc: Fixed in 84166eb8. With the status handling centralised in `GetKeyword`, `LoadKeyword` now owns the single failure notification and both callers just `return` on null, so the duplicated guard and comment are gone. Added `using Rock.Utility;` and dropped the two fully qualified `RockSerializableList` references.

PR #310, opened 2026-09-14
ROCK-9135 Use $(TargetDir) in plugin post-build events instead of hardcoded bin\Debug

stphnlee:
## Summary

Every legacy plugin csproj copied its build output into `RockWeb\bin` from a hardcoded `$(ProjectDir)bin\Debug\` path. Under a Release build that folder does not exist, `xcopy /R` exits 4, and the build fails. This blocks switching the Azure DevOps pipeline from Debug to Release.

This PR replaces the hardcoded prefix with `$(TargetDir)` so the copy follows the active configuration. It is a mechanical 1:1 substitution: 44 csproj files, 88 changed lines, nothing else touched.

- `org.secc.Workflow`: also reordered to match the sibling convention (org_secc folder first, then dll, then pdb) and added the missing `.pdb` copy.
- All other plugins: `$(ProjectDir)bin\Debug\` → `$(TargetDir)` only. Third-party DLL copies (iText, Ghostscript, PayPal, Owin, ImageResizer, SharpScss, etc.) come along automatically because they used the same prefix.

Debug builds are byte-for-byte unaffected: `$(TargetDir)` resolves to `bin\Debug\` in Debug.

## Verification

Built all 43 swept projects standalone in **Release** against the Rock checkout (`/p:SolutionDir=Rock\`), post-build events enabled:

| Pass | Projects | Succeeded | xcopy failures |
|------|----------|-----------|----------------|
| Release | 43 | 43 | 0 |
| Debug (restore) | 43 | 43 | 0 |

Release logs show every post-build copying from `bin\Release\`; no log line references `bin\Debug`. The Debug pass produced the same xcopy command count and the same set of "0 File(s) copied" lines as Release, so the only behavioral difference between configurations is the source folder.

`org.secc.Workflow` was additionally verified on its own before the sweep (Release build copied dll + pdb into RockWeb\Bin).

## Not in scope

- `BuildPlugin.ps1` scripts in a few plugin folders still reference `bin\Debug`. Those are manual `.plugin` package builders, not part of the solution build or the pipeline.
- Switching the pipeline itself to Release. That is a separate, deliberate change; this PR only removes the blocker.
- Consolidating the 44 hand-written xcopy blocks into a shared `Directory.Build.targets`. Worth doing later, after the pipeline is on MSBuild.

## Merge notes

Changes are confined to the `<PostBuildEvent>` block at the bottom of each csproj. Only open PR #304 touches one of these files (MetricsDigest) and in a different region; a rebase should be conflict-free.


jwakefield-secc (approved):
Looks good!

Issue #303, opened 2026-09-04
Workflow Launcher Plugin Incompatible With Rock v20


JimMichael:
While alpha testing Rock v20.0.9 we observed that the _Workflow Launcher_ plugin we rely on heavily throws an exception at startup:

<img width="2351" height="613" alt="Image" src="https://github.com/user-attachments/assets/7bdcc363-ec1d-4603-84a9-2fcb9889bc1f" />

stphnlee:
Thanks for letting us know. Which version of the plugin are you running? A text copy of the stack trace would help too. We'll dig in during our next sprint and post an update here by Sept 24.

JimMichael:
We're on Workflow Launcher plugin v1.4


(System.Web.HttpCompileException) D:\RockRMS\Plugins\org_secc\Workflow\WorkflowLauncher.ascx.cs(166): error CS1501: No overload for method 'LaunchWorkflow' takes 3 arguments

Exception Date
    9/4/2026 10:56 AM
Source
    System.Web

   at System.Web.Compilation.BuildManager.CompileWebFile(VirtualPath virtualPath)
   at System.Web.Compilation.BuildManager.GetVPathBuildResultInternal(VirtualPath virtualPath, Boolean noBuild, Boolean allowCrossApp, Boolean allowBuildInPrecompile, Boolean throwIfNotFound, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVPathBuildResultWithNoAssert(HttpContext context, VirtualPath virtualPath, Boolean noBuild, Boolean allowCrossApp, Boolean allowBuildInPrecompile, Boolean throwIfNotFound, Boolean ensureIsUpToDate)
   at System.Web.Compilation.BuildManager.GetVirtualPathObjectFactory(VirtualPath virtualPath, HttpContext context, Boolean allowCrossApp, Boolean throwIfNotFound)
   at System.Web.Compilation.BuildManager.GetCompiledType(VirtualPath virtualPath)
   at Rock.Web.Cache.BlockTypeCache.GetCompiledType() in \Rock\Web\Cache\Entities\BlockTypeCache.cs:line 300


Thanks!

stphnlee:
Please test if this version works for you on version 20 (GitHub wouldn't allow me to upload a .plugin file extension, so I renamed it to .zip. After downloading, you can change it back to .plugin).

Let me know if it works and I'll submit it to the Rock Shop.

Thanks!

[...]

PR #312, opened 2026-09-21
ROCK-9129 Make WorkflowEntityLaunch launch workflows on Rock v20 and v16

stphnlee:
## Summary

Makes the `WorkflowEntityLaunch` block actually launch workflows, on Rock v20 and v16. The block has never worked: its reflective `LaunchWorkflow` lookup never matched on any Rock version. This PR replaces the reflection with a direct, synchronous launch and hardens the block around it.

JIRA: [ROCK-9129](https://seccdev.atlassian.net/browse/ROCK-9129)
GitHub: secc/RockPlugins#303 (Workflow Launcher plugin incompatible with Rock v20)

## Root cause

- `WorkflowEntityLaunch.ascx.cs` looked up `Entity<T>.LaunchWorkflow` via `GetMethod` with `{ int, string, Dictionary<string,string> }`. The first parameter has been `int?` since 2016 and the plugin used `typeof( int )` since its first commit in 2019, so the lookup returned null on every Rock version and the Launch button silently did nothing.
- Rock v20 additionally removed the 3-argument `LaunchWorkflow` overloads; only the 4-argument `(int?/Guid?, string, Dictionary<string,string>, int?)` forms remain.
- The compile error reported in #303 (`WorkflowLauncher.ascx.cs(166)`, 3 arguments) does **not** reproduce against master. Master already passes 4 arguments (since ROCK-6271, July 2023). The Rock Shop package predates that change and needs to be republished from master.

## Change

`WorkflowEntityLaunch.ascx.cs` rewritten:

- **Direct launch.** Drops the `GetMethod`/`Invoke` lookup. Launches via `Workflow.Activate` + `WorkflowService.Process` inside a `RockContext`, wrapped in try/catch. Runs synchronously in the postback so the user sees real success or the actual error list instead of an optimistic "Launching..." message while a queued transaction fails silently. Acceptable tradeoff for a manual, single-entity button.
- **Initiator.** Sets `InitiatorPersonAliasId` to `CurrentPersonAliasId`, matching Rock core blocks. Previously always null. The sibling `WorkflowLauncher` block is deliberately left passing null; changing bulk launches is out of scope.
- **Null guards.** Workflow type, entity, `EntityTypeCache`/`GetEntityType`, `GroupMember.Group`, `Person.PrimaryAlias`. Missing entity or deleted workflow type now shows a message instead of a yellow screen on page load.
- **Inactive workflow type** is refused with a message on load and on click.
- **Duplicate-launch guard with override.** If an active workflow of the same type already exists for the entity, Launch lists the existing workflow(s) with links to Workflow Detail and does not launch. A **Launch Anyway** button then allows an intentional second launch (e.g. a stuck workflow). A replayed or double-submitted Launch postback still hits the guarded path.
- **HTML encoding.** Entity name, workflow type name, and error text are `EncodeHtml()`-ed before reaching the NotificationBox or Literal (stored XSS via Person names was possible).
- **Cleanup.** `getEntity()` collapsed to `Reflection.GetIEntityForEntityType`. Dead locals, unused `RockContext`, and `System.Reflection` import removed.

`WorkflowEntityLaunch.ascx`: `DataLoadingText="Launching..."` on the button; hidden `Launch Anyway` button.

`README.md`: updated to describe the new launch path.

## Behavior change to be aware of

Existing pages using this block have had a non-functional Launch button. After this ships, clicking it will launch workflows. Admins who have been clicking it with no effect should be told.

## Verification

- `org.secc.WorkflowLauncher.csproj` builds against Rock `20.0.9-alpha` (SparkDevNetwork/Rock `release-20.0`).
- `WorkflowEntityLaunch.ascx.cs` compile-checked with Roslyn `csc` against both Rock v20 and Rock v16 (`hotfix-1.16.12`) `RockWeb\Bin` assemblies; only the expected designer control-ID errors remain. These code-behinds are not in the csproj compile list and are compiled by ASP.NET at runtime, so the MSBuild pass alone does not cover them.
- Not yet exercised in a running RockWeb. Manual test: configure the block with a `GroupMember` entity type and a workflow type that maps `Group`/`Person` attributes, open with `?EntityId=`, click Launch, confirm the workflow exists with the expected attributes and initiator, click again and confirm the duplicate guard lists the first workflow and offers Launch Anyway, then confirm Launch Anyway creates a second one.

## Follow-up (not in this PR)

- Confirm installed Rock Shop version with the reporter.
- Build the plugin package from master against v20, publish to Rock Shop, reply on #303.
- Decide whether `WorkflowLauncher` (bulk block) should also pass the current person as initiator.

[ROCK-9129]: https://seccdev.atlassian.net/browse/ROCK-9129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwakefield-secc (approved):
Looks good! Checked the three Rock methods the block now calls against Rock source: unchanged in both our current version (1.16.12.1) and v20, so this should compile and behave the same after the upgrade. Guards, encoding, and error handling look correct. Blast radius is two staff-only pages (1812, 1961) under Camp Discount Code Reports, both on Registration, fed by one Dynamic Data link that already passes EntityId. The two workflow types finish in seconds, so running them in the postback is fine. Not compiled locally; please confirm a Windows build and smoke-test one Review click on DEV.

One question on the new "already active" guard: it refuses to launch if any open workflow of that type exists for the entity, with no override. No historical impact for the camp types, but is "one open at a time" the rule we want? Edge case: someone intentionally needs a second workflow for the same person or registration while the first is open or stuck. Would a short time window or a block setting to allow duplicates be a better fit?

stphnlee:
Thanks for the review, and good catch on the guard. Agreed that a hard "one open at a time" rule is too strict for a stuck workflow. Rather than a time window (arbitrary constant) or a block setting (an admin decision that the person at the page can't override), 051856a0 puts the decision in front of the user:

- **Launch** still refuses when an active workflow of the same type exists for the entity, but now lists each one (name, id, started, status, waiting-on activities) with a link to the core Workflow Detail page.
- A **Launch Anyway** button appears in that case and launches unconditionally.
- A replayed or double-submitted original Launch postback still hits the guarded path, so the accidental-duplicate protection is unchanged.

For the camp types this changes nothing in the normal path; it only adds an escape hatch when a previous run is open.

On verification: both `WorkflowEntityLaunch.ascx.cs` versions were compile-checked with Roslyn `csc` against the v16 (`hotfix-1.16.12`) and v20 `RockWeb\Bin` assemblies. I have not yet done a full Windows solution build or a DEV smoke test of a Review click; will do both before merging and report back here.

stphnlee:
Build and smoke test done on a local Rock **20.0.9** site (`release-20.0`, sample DB), commit 051856a0.

**Build**
- `org.secc.WorkflowLauncher.csproj` built with MSBuild against the Rock-20 `RockWeb\Bin` assemblies, no errors. Post-build deployed the dll and the four `.ascx` files.
- `WorkflowEntityLaunch.ascx.cs` compiled by ASP.NET on first page load with no errors.

**Smoke test** (block on a test page, EntityType = GroupMember, WorkflowType = the sample "Person Data Error" type, which stops at a user entry form so it stays active)

| Step | Result |
|---|---|
| Unconfigured block | "Please configure the block before using it." |
| No `EntityId` | "Please pass the EntityId as a page parameter." |
| `EntityId=99999` | "No entity of the configured type was found with ID 99999." |
| `EntityId=3` load | "Clicking Launch below will start a new instance of *Person Data Error* for *Admin Admin* (ID: 3)." |
| Launch | Workflow #1 created with `EntityTypeId`/`EntityId` set, `InitiatorPersonAliasId` = the logged-in user, active at the "Entry" activity |
| Launch again | Guard fired: lists workflow #1 (linked to `/Workflow/1`, started time, status, waiting-on activity) and shows **Launch Anyway** |
| Launch Anyway | Workflow #2 created, override button hidden again |
| Launch a third time | "2 active *Person Data Error* workflows already exist..." listing both |
| Exception Log | no rows during the test |

Not yet run against 1.16.12; the code-behind was compile-checked with `csc` against the v16 `RockWeb\Bin` but the v16 site shares the local IIS Express port with the v20 one, so that smoke test is still pending.

